PR #23, opened 2018-01-05
[travis] Test ROBOTOLOGY_USES_PYTHON option in Linux

traversaro:


traversaro:
The testing infrastructure changed completely in https://github.com/robotology/robotology-superbuild/pull/108 , and should have solved all the problems related to old SWIG version. I think we can close this PR.

PR #34, opened 2018-01-31
Add LD_LIBRARY_PATH in the setup.sh.in template

diegoferigo:
cc @traversaro


traversaro:
In theory, this should not be necessary in *nix if all the projects are correctly configured to use RPATH (and to ensure that @francesco-romano is happy : D ). 

However, if this is really necessary, I would suggest the following modifications: 
* I think this should be `DYLD_LIBRARY_PATH` on macOS (not really sure @claudiofantacci can probably help us on this) 
* it would be good to also document the use of `LD_LIBRARY_PATH` (and `DYLD_LIBRARY_PATH` in macOS) in the README to avoid that the setup.sh becomes an undocumented black hole

Just out of curiosity, what library is imposing a requirement on this?

claudiofantacci:
> I think this should be DYLD_LIBRARY_PATH on macOS (not really sure @claudiofantacci can probably help us on this)

I confirm that in macOS is `DYLD_LIBRARY_PATH`.

> In theory, this should not be necessary in *nix if all the projects are correctly configured to use RPATH (and to ensure that @francesco-romano is happy : D ).

I also agree on this, it should not be necessary and it should be avoided. `DYLD_LIBRARY_PATH` should be used to point to dynamic plugins folders that cannot be linked in any way.

diegoferigo:
In these days I cleaned my development setup and I started using this superbuild instead of the codyco one. I used to have also `LD_LIBRARY_PATH` in the exported variables, and not seeing it in the `setup.sh` let me think it was a lapse. I didn't remember about RPATH support.

This PR can be closed if `LD_LIBRARY_PATH` is not anymore required.

diegoferigo:
@francesco-romano Thanks for the explanation. Closing.

traversaro:
I think this variable is actually necessary for plugins loaded using `dlopen` such as RTF plugins, but let's add them properly when it is necessary.

PR #47, opened 2018-03-09
Just a check if issue 12 is actually solved

traversaro:


traversaro:
Travis is now using CMake 3.9, so this is quite difficult to reproduce, closing.

PR #54, opened 2018-03-14
Add macOs support in Travis CI

traversaro:
See https://github.com/robotology/gazebo-yarp-plugins/issues/351 .

PR #69, opened 2018-05-01
[travis] Enable macOS build

traversaro:


traversaro:
I forgot that I already started working on this in https://github.com/robotology/robotology-superbuild/pull/54 .

PR #105, opened 2018-08-03
Fix the behaviour when <PROJECT>_TAG is passed directly to CMake

diegoferigo:
This change prevents the superbuild to override user's variables

cc @traversaro


traversaro:
Cool. I guess cache variables may be more convenient (so the user can also edit them in the cache) but I am not sure about the implication. This will work for  now. 

cc @drdanz

PR #106, opened 2018-08-07
Updated WB-Toolbox informations

diegoferigo:
The toolbox's directory name inside `share/` was recently changed and the documentation needed to be updated accordingly. Furthermore, the tag has been bumped for including the latest bugfixes.

In order to avoid breaking controllers after https://github.com/robotology/wb-toolbox/pull/147, the tag in `master` points to the commit https://github.com/robotology/wb-toolbox/commit/0ff5ca4aecc607e1f1901abffe3b7c960a4e6b11.

Refer to https://github.com/robotology/wb-toolbox/pull/148.

### DIFF
--- a/README.md
+++ b/README.md
@@ -508,7 +508,7 @@ can install some projects that depend on MATLAB, in particular:
 
 To use this software, you can simply enable its compilation using the `ROBOTOLOGY_USES_MATLAB` CMake option.
 Once this software has been compiled by the superbuild, you just need to add some directories of the robotology-superbuild install (typically `$ROBOTOLOGY_SUPERBUILD_ROOT/build/install`) to [the MATLAB path](http://mathworks.com/help/matlab/matlab_env/add-folders-to-search-path-upon-startup-on-unix-or-macintosh.html).
-In particular you need to add to the MATLAB path the `$ROBOTOLOGY_SUPERBUILD_ROOT/build/install/mex` directory and all the subdirectories `$ROBOTOLOGY_SUPERBUILD_ROOT/build/install/share/WB-Toolbox`.
+In particular you need to add to the MATLAB path the `$ROBOTOLOGY_SUPERBUILD_ROOT/build/install/mex` directory and all the subdirectories `$ROBOTOLOGY_SUPERBUILD_ROOT/build/install/share/WBToolbox`.
 
 As an example, you could add this line to your MATLAB script that uses the robotology-superbuild matlab software:
 ~~~
--- a/cmake/ProjectsTagsMaster.cmake
+++ b/cmake/ProjectsTagsMaster.cmake
@@ -5,4 +5,4 @@ macro(set_tag tag_name tag_value)
 endmacro()
 
 set_tag(RTF_TAG d06f411e8c976018c62f520577a03085a4c300bf)
-set_tag(WBToolbox_TAG e7998ffc391f876a991bcd20f9d0eaf4266827f9)
+set_tag(WBToolbox_TAG 0ff5ca4aecc607e1f1901abffe3b7c960a4e6b11)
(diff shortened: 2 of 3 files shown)


traversaro:
I guess it needs also to be updated in https://github.com/robotology/robotology-superbuild/blob/master/cmake/template/startup_robotology_superbuild.m.in#L11 .

diegoferigo:
That script is installed by wb-toolbox in `<PREFIX>/share/WBToolbox/`. Probably removing it from the superbuild and tell the user to use directly the installed script is a more robust solution.

traversaro:
Honestly, I would prefer to avoid as much as possible the need for superbuild users to have to perform any manual project-specific installation step. See https://github.com/robotology/codyco-superbuild/pull/145 for the rational for which we added that script

A possible alternative (but I am not sure how  feasible it is) is to call the WB-Toolbox startup script from the superbuild one.

diegoferigo:
What I was proposing didn't add any additional step to the user. The current situation already requires the user to manually execute that script once, after the installation. I proposed to use the script shipped by wb-toolbox instead of the one copied-and-pasted in the superbuild.

However, I totally agree on automating the process. Calling that script from the superbuild would require executing a `matlab -nodesktop` instance, which is doable since matlab at that stage would be already found.

Note that this method works only if Matlab is executed from the [`userpath`](https://it.mathworks.com/help/matlab/matlab_env/assign-userpath-as-the-startup-folder-on-unix-or-macintosh.html) folder. This is always true for the icon launcher, not from the terminal. I personally always use the terminal method and for my usage the best option is using the env variable `MATLABPATH`.

Considering that we have the `setup.sh` and the two methods do not collide, we can provide support for both. What I don't know is if `matlab -nodesktop` works also on Windows.

traversaro:
> What I was proposing didn't add any additional step to the user. The current situation already requires the user to manually execute that script once, after the installation. I proposed to use the script shipped by wb-toolbox instead of the one copied-and-pasted in the superbuild.

Unfortunately, the superbuild `startup` script is still necessary, because adding the `<prefix>/mex` is necessary also for YARP/iDynTree Matlab bindings, and it is possible that the WB-Toolbox is not installed at all if the `ROBOTOLOGY_USES_MATLAB` option is used with the [`ROBOTOLOGY_NOT_USE_SIMULINK`](https://github.com/robotology/robotology-superbuild#system-dependencies-7) option (necessary if a users has  a Matlab license but not a Simulink one). 

> However, I totally agree on automating the process. Calling that script from the superbuild would require executing a matlab -nodesktop instance, which is doable since matlab at that stage would be already found.

We had several problems with that in the past. Unfortunately due to licensing issues, it is not uncommon that the matlab executable can only be launched by a user that is not the one building the superbuild. @gabrielenava had exactly this problem in the Walkman setup. 

> Considering that we have the setup.sh and the two methods do not collide, we can provide support for both. What I don't know is if matlab -nodesktop works also on Windows.

Using `MATLABPATH` you mean not using the startup script at all, right?

diegoferigo:
> Unfortunately, the superbuild startup script is still necessary, because adding the <prefix>/mex is necessary also for YARP/iDynTree Matlab bindings

You're right, bindings are installed in that folder too. That script is needed in any case.

> We had several problems with that in the past. Unfortunately due to licensing issues, it is not uncommon that the matlab executable can only be launched by a user that is not the one building the superbuild. @gabrielenava had exactly this problem in the Walkman setup.

Ow now I remember the script that launches the script. Yes, this is an edge case but unfortunately it might happen.

> Using MATLABPATH you mean not using the startup script at all, right?

Yes, I actually never used the script. The drawback is that this works only on systems where you can launch Matlab from the terminal, using the icon the environment is not loaded.

diegoferigo:
@traversaro For the time being let's stick to the current status. I edited the m script accordingly.

PR #117, opened 2018-09-13
Add testing against Gazebo in Travis

traversaro:


traversaro:
Actually it is still not working. : (

Issue #52, opened 2018-03-14
robotology-superbuild is not tested with Gazebo 9 installed from homebrew


traversaro:
Upstream issue: https://github.com/osrf/homebrew-simulation/issues/384 . 

For the time being, the recommended workaround is to install the `gazebo8` formula instead of the `gazebo9` one.

traversaro:
@robotology/all-robotology-developers If you are using homebrew and you need to use gazebo,  please use the **gazebo8** formula for the time being, thanks.

traversaro:
https://github.com/osrf/homebrew-simulation/issues/384 is now fixed, but I prefer to have a minimal CI setting testing this configuration before removing the warning added in https://github.com/robotology/robotology-superbuild/pull/53 .

PR #118, opened 2018-09-13
Add compilation of ROBOTOLOGY_USES_GAZEBO in Travis

traversaro:
Finally working on both macOS and Linux. : ) 
Fix https://github.com/robotology/robotology-superbuild/issues/52 .
 : )


traversaro:
Still too close to the 48-minutes limit, the net effect is that the test is flaky. Let's try to reduce of at least a minute the test, and then we eventually work to a more permanent solution (for example it seems to be that ccache si not properly working, and that will reduce the build time considerably).

diegoferigo:
@traversaro I did many attempts to use ccache in osx when using the Xcode generator but I never managed to make it work. If you use instead Unix Makefile it works.

traversaro:
> I did many attempts to use ccache in osx when using the Xcode generator but I never managed to make it work. If you use instead Unix Makefile it works. 

Cool, I missed this info.

traversaro:
The `brew upgrade` seems not to be strictly necessary, so I removed it for now. This saves ~4 minutes, that should be enough margin for now.

PR #126, opened 2018-10-11
Compile icub-firmware-shared when compiling robot

S-Dafarra:


traversaro:
This should be already handled in https://github.com/robotology/robotology-superbuild/blob/master/cmake/BuildICUB.cmake#L13 . It is not working properly?

PR #87, opened 2018-06-13
[WIP] Added files to add osqp, OsqpEigen, UnicyclePlanner and dcmWalking

S-Dafarra:
Right now it is not compiling since https://github.com/robotology/capture-point-walking does not have a ``CMakeLists.txt`` in the root folder.

cc @GiulioRomualdi

PR #129, opened 2018-10-20
[travis] Compile Dynamics profile based on env option

traversaro:
Combined with https://github.com/robotology/robotology-superbuild/commit/ff67ddb1d0ebc3891f54363ab59304f24b84805c that accidentally was committed on master, this would disable dynamics compilation on macOS, hopefully getting the mac build time inside the 50 minutes.


traversaro:
The latest master commit is building  again on Travis/macOS in less than 50 minutes  (44 minutes, check https://travis-ci.org/robotology/robotology-superbuild/jobs/444040909). I guess the actual amount of time is heavily dependent on which homebrew formula are available in bottle format. As new packages will be soon added to the `Dynamics` profile (see https://github.com/robotology/robotology-superbuild/pull/87) I think it  make sense to go forward and merge this PR.

PR #132, opened 2018-10-26
Update README.md

lrapetti:
A link in the README.md was no longer working, and I have improved the `addpath()` by using the `ROBOTOLOGY_SUPERBUILD_INSTALL_PREFIX ` and removing what looks a typo error (the word [`available`](https://github.com/robotology/robotology-superbuild/compare/master...lrapetti:patch-1#diff-04c6e90faac2675aa89e2176d2eec7d8L516) after the function).

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,16 +507,16 @@ can install some projects that depend on MATLAB, in particular:
 
 
 To use this software, you can simply enable its compilation using the `ROBOTOLOGY_USES_MATLAB` CMake option.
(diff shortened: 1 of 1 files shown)


traversaro:
Sorry, I just noticed a problem. If we actually want to use the `ROBOTOLOGY_SUPERBUILD_INSTALL_PREFIX`, we need to make sure that is defined also on Windows, and at the moment this is not mentioned in the docs. Can you add it in https://github.com/robotology/robotology-superbuild#configure-your-environment-2 ?

traversaro:
Thanks. We should eventually provide a windows-friendly variant of `setup.sh`. I did some tests in vcpkg-cpack (see https://github.com/robotology-playground/vcpkg-cpack/blob/master/addPathsToEnvVariables.ps1 and https://github.com/robotology-playground/vcpkg-cpack/blob/master/removePathsFromEnvVariables.ps1) and I think it should be doable, but I never had time to work on this. cc @claudiofantacci

PR #137, opened 2018-11-21
Add icub-gazebo-wholebody to ProjectsTagsDevel.cmake

gabrielenava:
I do not know if this is enough or if there is something else to modify @traversaro


traversaro:
It is enough, thanks!

Issue #134, opened 2018-11-12
Qt problems with latest Qt from vcpkg


traversaro:
@InesSorrentino experienced several problems related to running the **installed** version of `YARP` and `ICUB` GUIs that use QMLs when using the latest version of Qt installed via `vcpkg` . 

For some reason, the version of the executable contained in the `build` directory seem to be working fine. As there were recently some big changes in how Qt is installed by vcpkg (https://github.com/Microsoft/vcpkg/pull/4164 and https://github.com/Microsoft/vcpkg/pull/4629), it would be worth to check it this problems are actually reproducible.

claudiofantacci:
I guess they recompiled everything a was still not working. Am I right?
Speaking about QML and `vcpkg`, there is a specific folder in the vcpkg root folder that is specifically for QML things (not an expert of it, so I'm just throwing whatever comes into my mind here and there). Maybe we should have a look at it and understand whether we have to add folders in `PATH` or something similar.
Anyway, I'm still not 100% sure that the problem is related to `vcpkg` specifically.

InesSorrentino:
No, I am not using the vcpkg toolchain.

traversaro:
> I guess they recompiled everything a was still not working. Am I right?

Something changed: the executables in the build directory were working correct, but the installed were not working properly: in particular after the line `engine.load(QUrl(QStringLiteral("qrc:/main.qml")))` in https://github.com/robotology/icub-main/blob/4f5881acafab693f763dd277f87a8996fdea01a3/src/tools/iCubSkinGui-qt/src/main.cpp#L27 the `engine.rootObjects()` data structure had size 0, while it typically had size 1 in correctly working applications. Interesting, other applications like `yarpmotorgui` and `yarplogger` were instead working fine also from when launched from the installation directory. Even more interestingly, the `yarpamanager` application was not correctly loading the icons when launched from the installation directory, while everything was working fine (icon were visualized) when launched from the build directory.

>  Anyway, I'm still not 100% sure that the problem is related to vcpkg specifically.

I am trying to reproduce it on a second setup to be sure about it.

> If you are using the vcpkg toolchain, did you disable VCPKG_APPLOCAL_DEPS? This disables copying the dlls in the executable directory

[...]

PR #140, opened 2018-12-07
Remove instruction on using dependencies from vcpkg on Windows

traversaro:
Given https://github.com/robotology/robotology-superbuild/issues/134 , I think it make sense to do this, at least until https://github.com/robotology/robotology-superbuild/issues/134 is fixed.

cc @kouroshD


traversaro:
Travis failing due to this https://github.com/robotology/robotology-superbuild/issues/141, merging anyway.

Issue #145, opened 2018-12-12
Bug in OpenCV binaries distributed by YARP installer


traversaro:
See https://github.com/robotology/icub-packaging/issues/10 . 

@GiulioRomualdi

traversaro:
Thanks @Tobias-Fischer for reporting the issue.

traversaro:
We can close this after https://github.com/robotology/robotology-superbuild/pull/348 has been merged, as we will stop relying on those installers for opencv.

GiulioRomualdi:
Probably It may be close

PR #148, opened 2018-12-18
Document to install OpenCV from the official installers

traversaro:
Due to  https://github.com/robotology/robotology-superbuild/issues/145, we need to make sure that new Windows users of the superbuild are able to get a working installation, properly documenting the necessary steps.
As the installation is now unfortunately extremely complicated, I tried to change a bit the period to improve readability.


GiulioRomualdi:
This may be useful also to @InesSorrentino and @Yeshasvitvs

PR #152, opened 2018-12-20
Add a check before sourcing gazebo setup script

diegoferigo:
It might happen that the user configures and install the superbuild with gazebo support, and then uninstalls gazebo. Often, as referred in the `README`, users use the provided `setup.sh` script to configure the environment and they add it to their `.bashrc`.

This combination causes the `source` of the script to fail, and the fact that happens in the `bashrc` is a dangerous behavior. Luckily, in this case only a warning message like the follow is produced:

```
bash: /usr/share/gazebo/setup.sh: No such file or directory
```

This PR adds a check in order to prevent this behavior. Note that I couldn't use a regular `if` because the `;` before the `then` gets somehow removed by CMake after the template is instantiated.


traversaro:
> Note that I couldn't use a regular `if` because the `;` before the `then` gets somehow removed by CMake after the template is instantiated.

That is strange.

diegoferigo:
> That is strange.

I did not have time to investigate on this. It might be something related to the CMake function used for the template.

PR #156, opened 2019-01-02
Fix problem in configure_file_with_cmakeif when input file contains a semicolon

traversaro:
The problem was that various variables that contained part of the input file were used
in the CMake script unquoted, and so their content was interpreted as a semicolon-separated
CMake list, and not as a unique string. For more info, see point 3 of https://samthursfield.wordpress.com/2015/10/20/some-cmake-tips/ .

Originally reported by @diegoferigo in https://github.com/robotology/robotology-superbuild/pull/152 .


traversaro:
I am fully aware that the actual root of this is that `configure_file_with_cmakeif` is relavitly a huge hack, so if anyone has a better solution for the whole problem of generating `setup.sh` files feel free to propose it, thanks!

Issue #164, opened 2019-01-24
Enable YARP_USE_I2C when the superbuild is configured for the physical ICUB head 


traversaro:

traversaro:
This is necessary due to https://github.com/robotology/yarp/pull/1743 .

PR #165, opened 2019-01-24
Enable YARP_USE_I2C in YARP when ROBOTOLOGY_ENABLE_ICUB_ROBOT is true

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/164


nunoguedelha:
@traversaro we need the same change for `ENABLE_yarpmod_imuBosch_BNO055` right?

traversaro:
That one is already there: https://github.com/robotology/robotology-superbuild/pull/165/files#diff-63ade32be01d168eede4dfbb8f96a353L49 .

PR #172, opened 2019-02-02
Temporarily use v4 version of wb-toolbox

diegoferigo:
Waiting other repositories to align to the new version


diegoferigo:
:+1: I just did it but too late for this PR. It is a good reference in any case. The switch to the new wb-toolbox release might arrive quite soon, only the `whole-body-controllers` must be ported and the process is aided with a script. It should be straightforward.

PR #173, opened 2019-02-04
Pass ROBOTOLOGY_USES_MATLAB to wb-toolbox project

diegoferigo:


traversaro:
Thanks, mergin as MATLAB is not tested in CI.

PR #188, opened 2019-03-14
Add YCM publications to README.md

drdanz:


traversaro:
Can we put the DOI link as a normal text link instead of an image? A normal text simplifies the copy&pasting of the DOI, quite convenient if you manage your publications  using a program such as Zenodo or Mendeley.

drdanz:
Sure, do you prefer just the doi or or the doi url? (i.e. `DOI 10.1109/IRC.2018.00014` or `https://doi.org/10.1109/IRC.2018.00014`)

traversaro:
Both are ok, given that it is an URL probably the DOI URL makes more sense.

traversaro:
Thanks!

Issue #191, opened 2019-04-01
YARP shared library is linked against static system libSDL.a instead of libSDL.so 


VenusPasandi:
I tried to install the robotology-superbuild and I got the following error when I do `make`,

>[ 63%] Built target yarp_test_grabber
[ 63%] Linking CXX shared module ../../../lib/yarp/yarp_SDLJoypad.so
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libSDL.a(SDL.o): relocation R_X86_64_PC32 against symbol `SDL_InitSubSystem' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
src/devices/SDLJoypad/CMakeFiles/yarp_SDLJoypad.dir/build.make:125: recipe for target 'lib/yarp/yarp_SDLJoypad.so' failed
make[5]: *** [lib/yarp/yarp_SDLJoypad.so] Error 1
CMakeFiles/Makefile2:2417: recipe for target 'src/devices/SDLJoypad/CMakeFiles/yarp_SDLJoypad.dir/all' failed
make[4]: *** [src/devices/SDLJoypad/CMakeFiles/yarp_SDLJoypad.dir/all] Error 2
Makefile:151: recipe for target 'all' failed
make[3]: *** [all] Error 2
CMakeFiles/YARP.dir/build.make:108: recipe for target 'robotology/YARP/CMakeFiles/YCMStamp/YARP-build' failed
make[2]: *** [robotology/YARP/CMakeFiles/YCMStamp/YARP-build] Error 2
CMakeFiles/Makefile2:68: recipe for target 'CMakeFiles/YARP.dir/all' failed
make[1]: *** [CMakeFiles/YARP.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2

I don't know what is the reason but I got the following warning when I did `cmake` 

> CMake Warning (dev) at build/CMakeFiles/YCMEPHelper.cmake:171 (message):
>    YCM_BOOTSTRAP HASH mismatch
> 
>      for file: [/home/vpasandi/Documents/repositories/robotology-superbuild/cmake/IncludeUrl.cmake]
>        expected hash: [e31dffed0897729283ef4e945bbdeb1921b438b5]
>          actual hash: [ae6de83398115e5b8fd6c261972dea82d1aeb3fd]
> 
[...]

PR #193, opened 2019-04-01
Update YCM's IncludeUrl.cmake to latest version

traversaro:
IncludeUrl.cmake is a file vendored  to permit to bootstrap YCM, so it needs to be updated whenever it is changed in YCM (see http://robotology.github.io/ycm/gh-pages/v0.10/manual/ycm-using.7.html#using-ycm-as-soft-dependency). 
Get version from commit https://github.com/robotology/ycm/commit/3dcadbd146ad5c0cdba7034c696bc73b11b4a5a0,
released as part of YCM 0.10 .

Solve the warning mentioned in https://github.com/robotology/robotology-superbuild/issues/191

### DIFF
--- a/cmake/IncludeUrl.cmake
+++ b/cmake/IncludeUrl.cmake
@@ -165,7 +165,7 @@ macro(INCLUDE_URL _remoteFile)
 
   get_filename_component(_filename ${_remoteFile} NAME)
   if(DEFINED _IU_DESTINATION)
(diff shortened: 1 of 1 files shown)


VenusPasandi:
Sorry for the delay, apparently @traversaro merged the changes to the master. I tried the master branch and now there is no warning about YCM update.

PR #196, opened 2019-04-03
Enable ccache in linux travis jobs

### LINKED ISSUE #195: Enable ccache in Travis CI for linux jobs
At a first glance, it doesn't seem that we exploit ccache in our linux jobs running in Travis CI. The folder is mounted:

https://github.com/robotology/robotology-superbuild/blob/288b636fdc3498403bcbdc6f6c9d32d300cc7759/.travis.yml#L38-L49

 but inside the container 1) [there is no ccache](https://github.com/robotology/robotology-superbuild/blob/master/.ci/install_debian.sh) and 2) is not enabled ([how to enable ccache](https://manpages.ubuntu.com/manpages/xenial/man1/ccache.1.html)). Addressing it might solve most of the timeouts we experience.

cc @traversaro

diegoferigo:
Closes #195

### DIFF
--- a/.ci/install_debian.sh
+++ b/.ci/install_debian.sh
@@ -7,7 +7,7 @@ apt-get update
 export DEBIAN_FRONTEND=noninteractive
 
 # CI specific packages
-apt-get install -y clang valgrind
+apt-get install -y clang valgrind ccache
 
 # Core dependencies
 apt-get install -y build-essential cmake clang coinor-libipopt-dev libboost-system-dev libboost-filesystem-dev libboost-thread-dev libeigen3-dev libtinyxml-dev libace-dev libgsl0-dev libopencv-dev libode-dev liblua5.1-dev lua5.1 git swig qtbase5-dev qtdeclarative5-dev qtmultimedia5-dev libqt5opengl5-dev libqcustomplot-dev libxml2-dev
--- a/.ci/install_debian_and_script.sh
+++ b/.ci/install_debian_and_script.sh
@@ -3,5 +3,8 @@ set -e
 
 DIR=$(dirname "$(readlink -f "$0")")
 
+# Enable ccache
+export PATH=/usr/lib/ccache:$PATH
+
 sh $DIR/install_debian.sh
 sh $DIR/script.sh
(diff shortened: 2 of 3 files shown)


diegoferigo:
Let's wait CI. I suspect that in order to have the ccache folder populated and its archive uploaded, the build should succeed. It is like a dog biting its tail. Let's see what happens.

diegoferigo:
Only the last one timed out. I restart the build, let's compare the speedup now that ccache was populated.

Screenshot_20190403_182742

traversaro:
It does not seem to be faster, but perhaps it needs to be merged in master before. If you agree, we can merge and see if the modifications is useful.

diegoferigo:
@traversaro For some reason using $HOME/.ccache as target destination from root does not work. Explicitly using `/root/.ccache` works, and the cache is both correctly created and mounted for later use.

This means that the run you compared with the previous screenshot did not have yet ccache properly enabled. See this one that comes from yesterday's build of my fork:

Screenshot_20190404_092543

Waaay better :)

diegoferigo:
I don't know why that job always times out. Other jobs can reach 40 minutes easily, but that one is always killed before reaching even 30.

traversaro:
> I don't know why that job always times out. Other jobs can reach 40 minutes easily, but that one is always killed before reaching even 30.

I suspect they silently introduced some cap on the total time of the all jobs. We already observed some weird behavior, such as periodic daily builds switched to monthly, so this does not surprise me. As most the linux users (with your exception, to be honest :D ) do not use clang for compiling, I think it could make sense to just drop Linux clang builds, even considering that most projects have their own CI.

diegoferigo:
> I think it could make sense to just drop Linux clang builds

I agree, let's see if the the situation is solved once the cache kicks in (in my fork it doesn't seem it did, as you can see in the screenshot).

traversaro:
Can we merge in the mean time? For sure your modifications are an improvements, and the CI is broken anyway at the moment, so I think we can just merge it as it is.

traversaro:
The problem persists, let's merge and disable clang.

diegoferigo:
Ack

diegoferigo:
I had to squash the commit, too late :sweat_smile:

Issue #189, opened 2019-03-22
Cleanup YCM use after YCM 0.10 release


traversaro:
* [ ] Require YCM 0.10 after its release 
* [x] Update YCMBootstrap and IncludeUrl to latest version (after the fix in https://github.com/robotology/ycm/issues/93)
* [x] Remove the `YCM_SKIP_HASH_CHECK` option set to ON

traversaro:
The last two points have been already address in https://github.com/robotology/robotology-superbuild/pull/175 .

Issue #21, opened 2017-12-20
Specify a desired version in YCMBootstrap 


traversaro:
In the existing superbuild, we routinely change the `YCM_TAG` option to specify that we require some newer version of ycm to build the superbuild. However, when we check if a YCM is available in the system, we do not  check for its version, so it often happens that the superbuild is not working because the user has an outdated ycm installed in his system. 

This is similar to the problem that applies to generic projects in the superbuild: https://github.com/robotology/codyco-superbuild/issues/161 .

drdanz:
If I remember correctly, YCM version for development builds contains the date information, therefore we should be able two use that for checking the version. An alternative is to increase the patch version for the development branch more often...

traversaro:
The point I was making it is a bit different. Now that you released YCM v0.6.0, I can switch to it in  https://github.com/robotology/robotology-superbuild/blob/master/CMakeLists.txt#L93 , but I still can't set the minimum desired version in https://github.com/robotology/robotology-superbuild/blob/master/cmake/YCMBootstrap.cmake#L74 . 
I could fix this by modifying the YCMBootstrap file, but this is against all the hash checks.

PR #199, opened 2019-05-03
Use v0.10.2 of YCM

traversaro:
The exact version if  using the bootstrapped YCM, or at least that version when using system's YCM

Fix https://github.com/robotology/robotology-superbuild/issues/21 
Fix https://github.com/robotology/robotology-superbuild/issues/189
Fix https://github.com/robotology/robotology-superbuild/issues/191


traversaro:
I also updated the minimum required version of CMake to 3.5, as that is the minimum required version by YCM v0.10

traversaro:
Build failure is related to https://github.com/robotology/ycm/issues/254 .

PR #203, opened 2019-05-06
travis: add --network=host option to docker

traversaro:
Workaround for  https://github.com/robotology/ycm/issues/254

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_script:
 script:
   - echo "max_size = 1.0G" > $HOME/.ccache/ccache.conf 
   - >-
-    docker run -it \
+    docker run --network=host -it \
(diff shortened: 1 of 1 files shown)


diegoferigo:
To add some context to this PR, by default Docker creates a standalone network for the container (with a different hostname and ip). Using network host means that the container uses the very same network (and hence, ip and hostname) of the host OS.

If there's only one container running, as in this case, it is not a big difference. Though, host network mode is incompatible with running multiple containers that use the same port due to collisions.

traversaro:
Apparently things are starting to work, let's wait this evening to check if we can avoid this, otherwise we will merge.

traversaro:
The issue seems to be resolved at the Travis level, see https://www.traviscistatus.com/incidents/kyf149kl6bvp . Declining the PR as the less strange configuration we have, the better.

PR #207, opened 2019-05-31
Update README.md

bemilio:
From how it was previously written, I didn't understand that line had to go in the bashrc and that gave me problems. I think it is clearer now

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ Software installed by the following [profile](#profile-cmake-options) or [depend
 
 
 As a convenient feature the superbuild provides an automatically generated `setup.sh` sh script that will set
-all the necessary enviromental variables:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,7 +148,7 @@ Software installed by the following [profile](#profile-cmake-options) or [depend
 
 
 As a convenient feature the superbuild provides an automatically generated `setup.sh` sh script that will set
-all the necessary enviromental variables:
+all the necessary enviromental variables. To do so append the following line to the .bashrc file:
```
traversaro:
```suggestion
all the necessary enviromental variables. To do so append the following line to the `.bashrc` file:
```

traversaro:
Thanks, this makes definitely more sense!

PR #211, opened 2019-06-10
Update README to install ipopt from homebrew core

traversaro:
See dartsim/homebrew-dart#76 for a related discussion.

@lrapetti @nunoguedelha Can you check if the ipopt formula from homebrew core  is working correctly? Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,9 +165,9 @@ user@host:~$ sudo ldconfig
 ## macOS
 
 ### System Dependencies
-To install Eigen and CMake, it is possible to use [Homebrew](http://brew.sh/):
+To install the system dependencies, it is possible to use [Homebrew](http://brew.sh/):
 ```
-brew install ace eigen cmake boost tinyxml swig qt5 gsl pkg-config jpeg sqlite readline tinyxml dartsim/dart/ipopt
+brew install ace eigen cmake boost tinyxml swig qt5 gsl pkg-config jpeg sqlite readline tinyxml ipopt
 ```
 
 Since Qt5 is not symlinked in `/usr/local` by default in the homebrew formula, `Qt5_DIR` needs to be properly set to make sure that CMake-based projects are able to find Qt5.


lrapetti:
I tried to uninstall the `dartsim/dart/ipopt` to get the one from homebrew core, but after I did that something broke I think because of the version of the dependencies.

#### opencv (solved)
In particular, if I try to run the `yarpdataplayer` I was getting the following error:
```
dyld: Library not loaded: /usr/local/opt/openblas/lib/libopenblasp-r0.3.5.dylib
  Referenced from: /usr/local/opt/opencv@3/lib/libopencv_core.3.4.dylib
  Reason: image not found
Abort trap: 6
```
so I had to reinstall opencv@3 via homebrew and the problem with the yarpdataplayer was fixed.

####  iDynTree
I tried then to recompile iDynTree changing the path from `/usr/local/Cellar/ipopt/3.12.9/lib/libipopt.dylib` to `/usr/local/Cellar/ipopt/3.12.12/lib/libipopt.dylib` (I had an older version on my machine), but when compiling the headers are not found:
```
/include/private/InverseKinematicsNLP.h:14:10: fatal error: 
      'IpTNLP.hpp' file not found
#include <IpTNLP.hpp>
         ^~~~~~~~~~~~
```

traversaro:
> I tried then to recompile iDynTree changing the path (..)

I am afraid this is due to the fact that `IPOPT_INCLUDE_DIR` is still pointing to the old include dirs, as it is a CACHE variable that is not updated by https://github.com/robotology/idyntree/blob/master/cmake/FindIPOPT.cmake#L56 even if `_PC_IPOPT_INCLUDE_DIRS` changes.  Can you try to remove from the `build/robotology/iDynTree/CMakeCache.txt` all the ipopt-related variables and retry?

lrapetti:
> > I tried then to recompile iDynTree changing the path (..)
> 
> I am afraid this is due to the fact that `IPOPT_INCLUDE_DIR` is still pointing to the old include dirs, as it is a CACHE variable that is not updated by https://github.com/robotology/idyntree/blob/master/cmake/FindIPOPT.cmake#L56 even if `_PC_IPOPT_INCLUDE_DIRS` changes. Can you try to remove from the `build/robotology/iDynTree/CMakeCache.txt` all the ipopt-related variables and retry?

Thanks, I confirm that after this change it compiles, and I was able to run the code using Ipopt.

PR #212, opened 2019-06-12
Create some documentation related to use the robotology-superbuild on the icub-head/pc104

traversaro:
Basically I am just moving some info from https://github.com/robotology/robotology-superbuild/issues/88 to a more permanent location. 

Note that this is not meant to be a duplicate of  the information provided in the iCub wiki, as iCub wiki pages to describe the installation with the superbuild are already provided, and are also linked in the new document.

PR #214, opened 2019-06-21
Add missing ICUB_HEAD profile options in the BuildICUB script

traversaro:
As emerged during the tests on iCubGenova01 ( see https://github.com/robotology/robotology-superbuild/pull/212 ).


traversaro:
I will merge without waiting for CI as the ICUB_HEAD profile is not tested in the CI.

Issue #68, opened 2018-04-21
Cleanup usage of CMAKE_ARGS and CMAKE_CACHE_ARGS in Build**.cmake files


traversaro:
It is necessary to cleanup the usage of the `CMAKE_ARGS`, `CMAKE_CACHE_ARGS` and  `CMAKE_DEFAULT_CACHE_ARGS` in the calls to `ycm_ep_helper` in `Build***.cmake` files. 

Back in ~2014, I remember that we always had  `CMAKE_CACHE_ARGS`, but then we started adding subproject CMake options that depended on superbuild options, and so we had problem when switching this option on and off that the relative change was not correctly propagated to the subproject (as the initial cache was already populated, see https://github.com/robotology/ycm/issues/22 and https://github.com/robotology/codyco-superbuild/issues/5). From that point, we basically randomly used  `CMAKE_ARGS` and `CMAKE_CACHE_ARGS` until we reached a configuration that was working for most users. 

Given the discussion in https://github.com/robotology/robotology-superbuild/issues/67#issuecomment-383281344 and in  https://github.com/robotology/robotology-superbuild/issues/68#issuecomment-386028343, I think it make sense to review all existing `Build***.cmake` files to make sure that:
* `CMAKE_CACHE_DEFAULT_ARGS`  can be used with the ExternalProject fork in YCM, 
* `CMAKE_ARGS` and `CMAKE_CACHE_ARGS`  or  is only used for subproject options that depend on superbuild options, 
* `CMAKE_CACHE_DEFAULT_ARGS` is used for all other variables 

A more corner-case but still typical use of  `CMAKE_ARGS` is to make sure that a new modification to the `Build***.cmake`  is also propagated correctly to all existing superbuild installations.

traversaro:
cc @diegoferigo

drdanz:
I'm don't remember if the `ExternalProject` version in YCM supports it, but there is also the `CMAKE_CACHE_DEFAULT_ARGS` option, that should be used for cache arguments that could be changed by the user...

traversaro:
Note: in https://github.com/robotology/robotology-superbuild/pull/76 I added the option `-DICUB_INSTALL_WITH_RPATH:BOOL=ON` to `CMAKE_ARGS` because I wanted it to be propagated also to existing builds. However, once some time have passed, I think we can migrate it to `CMAKE_CACHE_ARGS` .

traversaro:
[...]

Issue #67, opened 2018-04-20
Document how to handle option of a subproject not handled by the superbuild


GiulioRomualdi:
I'm using the `robotology-superbuild` to install YARP on Ubuntu 16.04 but I do not find a way to enable internal YARP options through the superbuild (e.g. `CREATE_DEVICE_LIBRARY_MODULES` and `ENABLE_yarpmod_SDLJoypad`)

Is there a way to enable these options by using the superbuild or should I install YARP following the [standard instructions](http://www.yarp.it/install_yarp_linux.html)?

traversaro:
The workflow that we typically use for these cases is:
* Build the superbuild as usual 
* Once you have a valid build of the YARP subproject, `cd` in the build directory of the YARP subproject and configure it, using `cmake`, `ccmake` or `cmake-gui`. 
* Enable the options you are interested in, such as `ENABLE_yarpmod_SDLJoypad`. As this point, the option will be saved in the [CMake cache](http://cgold.readthedocs.io/en/latest/tutorials/variables/cache.html), and so they will be respected also if you compile the superbuild as an whole. 
* Note that this strategy will not work for variables that are explicitly set in the [`CMAKE_ARGS`](https://github.com/robotology/robotology-superbuild/blob/e4fa0621db9100e7958be73de28ed950214145e3/cmake/BuildYARP.cmake#L30) argument in `BuildYARP.cmake`. This is because the `CMAKE_ARGS` are passed to the cmake command used for configuring the YARP subproject **every  time** the subproject is configured, overwriting (if present) any value written in cache that is also listed in the `CMAKE_ARGS` (see the [`ExternalProject` module docs](https://cmake.org/cmake/help/latest/module/ExternalProject.html) for more info on this) . Actually this can be easily avoided, I will open an issue on this. 

As you may imagine, this work-flow is not really intuitive/user friendly, so it may be worth to properly document it. Note that this approach is not useful when you want to toggle options that change the dependencies of the project, as discussed for example in https://github.com/robotology/robotology-superbuild/issues/1#issuecomment-351773847 . 

[...]

PR #216, opened 2019-06-24
If ICUB_HEAD profile is enabled enable serialport device in YARP

traversaro:
This was detected by @pattacini when testing face expressions on the `iCubGenova01` "Blacky" robot. This matches indeed the documentation present in http://wiki.icub.org/wiki/Compilation_on_the_pc104 . In that page, regarding YARP also the `serial` device was enabled, but apparently the `serial` device is a builtin device of YARP (https://github.com/robotology/yarp/blob/v3.1.2/src/libYARP_dev/src/devices/ServerSerial/ServerSerial.cpp#L17) so there is no need to explicitly enable it (probably this changed at some point in the past). Furthermore, also the `portaudio` device  is enabled, but I do not see it used anywhere in https://github.com/robotology/robots-configuration/search?q=portaudio&unscoped_q=portaudio, so I guess that device is not actually required.

### DIFF
--- a/cmake/BuildYARP.cmake
+++ b/cmake/BuildYARP.cmake
@@ -44,6 +44,7 @@ ycm_ep_helper(YARP TYPE GIT
                               -DENABLE_yarpmod_fakebot:BOOL=ON
                               -DENABLE_yarpmod_imuBosch_BNO055:BOOL=${ROBOTOLOGY_ENABLE_ICUB_HEAD}
                               -DENABLE_yarpmod_SDLJoypad:BOOL=ON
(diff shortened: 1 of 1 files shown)


traversaro:
Note that due to https://github.com/robotology/robotology-superbuild/issues/68, this options could create problems if someone has `ROBOTOLOGY_ENABLE_ICUB_HEAD` set to `OFF`, but wants to set to `ON` the `ENABLE_yarpmod_serialport` YARP option. I think that fixing this problem however is outside the scope of this PR. (Another related issue: https://github.com/robotology/robotology-superbuild/issues/67)

PR #224, opened 2019-07-01
Bump YCM bootstrapped version to latest released version v0.10.4

traversaro:
Without this change, YARP downloads its own version of YCM, as 0.10.2 is not recent enough for YARP.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,8 +99,8 @@ endif()
 # Bootstrap YCM
 set(YCM_FOLDER robotology)
 set(YCM_COMPONENT core)
-set(YCM_TAG v0.10.2)
-set(YCM_MINIMUM_VERSION 0.10.2)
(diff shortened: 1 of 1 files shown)


traversaro:
Eventually we can just track the latest `master` of YCM as we do for most `robotology` dependencies, but historically we used fixed release for YCM as the boostrap process is quite delicate.

traversaro:
> If there are no strong opinions against using a specific version of YCM, I would just switch to master in order to simplify maintenance

That's indeed what we always did with the rest of robotology-software. At some point in time I stopped tracking master for YCM because the bootstrap mechanism is quite delicate and tended to break quite frequently. I think we can now switch to `master` again, but let's think a bit about it.

traversaro:
Test failures are unrelated (actually due to https://github.com/robotology/gazebo-yarp-plugins/issues/433), merging.

PR #230, opened 2019-07-26
Fix the export of PYTHONPATH env variable

diegoferigo:
As reported in https://github.com/robotology/robotology-superbuild/pull/82#discussion_r307357684, the `export` keyword was missing.

cc @HosameldinMohamed


traversaro:
Thanks to @diegoferigo and @HosameldinMohamed !

PR #234, opened 2019-08-22
Add GitHub Actions-based Continuous Integration for Ubuntu 18.04

traversaro:


diegoferigo:
Have a look at https://github.com/robotology/how-to-export-cpp-library/pull/47 I just submitted for multiplatform support :)

traversaro:
This extensive CI already found a bug, `ROBOTOLOGY_ENABLE_ICUB_HEAD` does not work on macOS due to the dragonfly option. : )

diegoferigo:
You can also enlarge the test matrix and include / exclude jobs [as documented here](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix).

traversaro:
I think the first version is ready to be merged, as it already provides testing for iCub Head and Human Dynamics profile, that are not currently tested in the Travis CI. 
Further work will involve:
* Fixing macOS ( https://github.com/robotology/robotology-superbuild/issues/237 )
* Fixing compilation with `ROBOTOLOGY_USES_GAZEBO` ( https://github.com/robotology/robotology-superbuild/issues/237 )

traversaro:
@diegoferigo This is now ready for review. 
* Do not worry about the number of commits, I will squash and merge from the GitHub UI.
* There are still some mac-related stuff, that I left because we will use them as soon as we fix https://github.com/robotology/robotology-superbuild/issues/237 , so I did not remove them, as they are simply inactive if `macOS-latest` is not in the `os` array.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,69 @@
+name: C++ CI Workflow
+
+on:
+  push:
+    branches:
+      - master
```
diegoferigo: Is there any reason to limit this workflow to run only on master? What about new branches upstream / new branches on a fork? I'm not sure if the pipeline is triggered anyways in this case.

traversaro: This are good points, but my rationale is the following: robotology-superbuilds jobs will be quite heavy, and when you create a new commit you are typically working on docs or CMake structure, and so I think it make sense to defer running the actual CI to when a new PR is opened. We can always revise this in the future.

diegoferigo: Ok, good to know :+1:

PR #249, opened 2019-09-06
Windows environment configuration - docs error

AlbertoRemusIIT:
"suites" instead of "suits" at line 350


traversaro:
Good catch! For reference, the related change was done in icub-tests in https://github.com/robotology/icub-tests/pull/47 . 
Unfortunately that info about what should be part of `YARP_DATA_DIRS` is a bit duplicate among OS-specific documentation and automatic scripts, see https://github.com/robotology/robotology-superbuild/search?q=suits&unscoped_q=suits . Could you update all the instances of `suits` with `suites`? Thanks!

AlbertoRemusIIT:
I did one pull request for each file, was it possible to do it all together?

traversaro:
> got it, I still have to modify the other 4 instances right?

Yes, if the changes are all related to the same PR even if in different files, you can just modify directly in the branch of the PR (in this case https://github.com/AlbertoRemusIIT/robotology-superbuild/tree/patch-1) without the need to create 4 different PRs.

AlbertoRemusIIT:
> Yes, if the changes are all related to the same PR even if in different files, you can just modify directly in the branch of the PR (in this case https://github.com/AlbertoRemusIIT/robotology-superbuild/tree/patch-1) without the need to create 4 different PRs.

Understood

traversaro:
Friendly ping @AlbertoRemusIIT (this does not have high priority, just checking you did not forgot : ) ).

traversaro:
It seems that modifications for `removePathsFromUserEnvVariables.ps1.in`, `setup.sh.in` and `setup.bat.in` are still missing.

AlbertoRemusIIT:
> Yes, if the changes are all related to the same PR even if in different files, you can just modify directly in the branch of the PR (in this case https://github.com/AlbertoRemusIIT/robotology-superbuild/tree/patch-1) without the need to create 4 different PRs.

If I remember well they were in other branches

traversaro:
Thanks!

Issue #169, opened 2019-01-29
Document that either opencv@2 or opencv@3 need to be installed in homebrew


nunoguedelha:
### Failure description
The YARP sub-project build fails with the following `Cmake` options:
```
ROBOTOLOGY_ENABLE_CORE
ROBOTOLOGY_ENABLE_DYNAMICS
ROBOTOLOGY_USES_GAZEBO

 YCM_EP_DEVEL_MODE_BlockFactory   OFF                                                                                                 
 YCM_EP_DEVEL_MODE_GazeboYARPPl   ON                                                                                                  
YCM_EP_DEVEL_MODE_ICUB           ON                                                                                                  
 YCM_EP_DEVEL_MODE_ICUBcontrib    OFF                                                                                                 
 YCM_EP_DEVEL_MODE_RTF            ON                                                                                                  
 YCM_EP_DEVEL_MODE_WBToolbox      ON                                                                                                  
 YCM_EP_DEVEL_MODE_YARP           ON                                                                                                  
 YCM_EP_DEVEL_MODE_YCM            OFF                                                                                                 
 YCM_EP_DEVEL_MODE_codyco-modul   ON                                                                                                  
 YCM_EP_DEVEL_MODE_iDynTree       ON                                                                                                  
 YCM_EP_DEVEL_MODE_icub-gazebo    OFF                                                                                                 
 YCM_EP_DEVEL_MODE_icub-gazebo-   ON                                                                                                  
 YCM_EP_DEVEL_MODE_icub-models    OFF                                                                                                 
 YCM_EP_DEVEL_MODE_icub-tests     ON                                                                                                  
 YCM_EP_DEVEL_MODE_whole-body-c   OFF                                                                                                 
 YCM_EP_DEVEL_MODE_wholeBodyInt   OFF                                                                                                 
 YCM_EP_DEVEL_MODE_yarp-matlab-   ON                                                                                                  
 YCM_EP_DEVEL_MODE_yarpWholeBod   OFF                                                                                                 
```
after updating all the repos on the 29/01/2019, and checkout the `devel` branch on all the sub-projects which are in `DEVEL_MODE`.

### Details on the configuration and error message
Error message:
```
[...]

PR #255, opened 2019-09-18
Update README with opencv details for MacOS

lrapetti:
Related discussion in https://github.com/robotology/robotology-superbuild/issues/169.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Since Qt5 is not symlinked in `/usr/local` by default in the homebrew formula, `
 export Qt5_DIR=/usr/local/opt/qt5/lib/cmake/Qt5
 ```
 
-The required version of `opencv` is not the latest release. In case other versions are installed, it is required to link `opencv3`.
+The version of `opencv` required by [YARP is not the latest release](https://github.com/robotology/yarp/issues/1672). In case other versions are installed, it is required to link `opencv3`.
 ```
 brew link --force opencv@3
 ```
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ The version of `opencv` required by [YARP is not the latest release](https://git
 ```
 brew link --force opencv@3
 ```
-Otherwise, it is possible to not use `opencv` by setting `YARP_USES_OPENCV` `OFF`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -175,14 +175,20 @@ user@host:~$ sudo ldconfig
...
 Since Qt5 is not symlinked in `/usr/local` by default in the homebrew formula, `Qt5_DIR` needs to be properly set to make sure that CMake-based projects are able to find Qt5.
 ```
 export Qt5_DIR=/usr/local/opt/qt5/lib/cmake/Qt5
 ```
 
+The required version of `opencv` is not the latest release. In case other versions are installed, it is required to link `opencv3`.
+```
+brew link --force opencv@3
+```
+Otherwise, it is possible to not use `opencv` by setting `YARP_USES_OPENCV` `OFF`.
```
traversaro:
```suggestion
Otherwise, it is possible to not use `opencv` by setting the `YARP_USE_OpenCV` YARP CMake option to  `OFF`. To modify YARP-specific CMAke options, you need to go with the terminal to `build/robotology/YARP`, and use the `ccmake` command to edit the YARP CMake options. 
```

In my experience, average superbuild users can be easily confused between CMake options of the superbuild and the one of the subprojects, so it is better to be clear. I also modified the YARP option that has a slightly different spelling.

lrapetti: > average superbuild users can be easily confused between CMake options of the superbuild and the one of the subprojects, so it is better to be clear

The options seem to be visible only if toggled advanced mode is ON, do you think it is better to mention also that?

traversaro: I added an "advanced" adjective to the suggest phrase, let's hope that this is enough.

lrapetti: ok, thanks @traversaro. I have merged your commit suggestions.

traversaro:
Thanks @lrapetti !

traversaro:
I noticed that in https://github.com/robotology/robotology-superbuild/pull/338/checks?check_run_id=417957153 the build was failing because OpenCV was not found, and I found  that apparently `opencv@3` is keg-only, and to actually use it you need to set  `OpenCV_DIR` to `/usr/local/opt/opencv@3/share/OpenCV` . In  theory we should update the docs added in this PR, but as we will be soon migrate to OpenCV 4 ( https://github.com/robotology/robotology-superbuild/issues/324 ) and I guess that improving now the docs for opencv@3 is not strictly required.

PR #258, opened 2019-09-19
Create regression test to reproduce issue 174 

traversaro:
See https://github.com/robotology/robotology-superbuild/issues/174 for more details. 
cc @lrapetti

### DIFF
--- a/.github/workflows/174-regression-test.yml
+++ b/.github/workflows/174-regression-test.yml
@@ -56,4 +56,5 @@ jobs:
         cmake -DROBOTOLOGY_ENABLE_DYNAMICS:BOOL=ON ..
         make 
         # Verify that YARP is a still a subproject
+        make YARP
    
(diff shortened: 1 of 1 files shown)


.github/workflows/174-regression-test.yml
```diff
@@ -0,0 +1,59 @@
...
+        # Normal build
+        cmake ..
+        make
+        # Verify that YARP is a subproject
+        make YARP
+        # Now enable ROBOTOLOGY_ENABLE_DYNAMICS
+        cmake -DROBOTOLOGY_ENABLE_DYNAMICS:BOOL=ON ..
+        make 
+        # Verify that YARP is a still a subproject
+
```
lrapetti: @traversaro Isn't `make YARP` missing here?

traversaro: Yes, thanks for checking

traversaro:
The attempt of reproducing the the issue #174 failed.

Issue #101, opened 2018-07-26
Add a CI (GitHub Actions) job building projects in unstable branch


traversaro:
This can be done with https://github.com/robotology/robotology-superbuild/pull/100 . 
~~Related issue: https://github.com/robotology/robotology-superbuild/issues/96 .~~

traversaro:
>  Related issue: #96 .

Things has changed, I think now it makes more sense to set it up with GitHub Actions.

diegoferigo:
If the tests will be executed in a clean docker image by building a given Dockerfile, in order to test both devel and master with the same file you can refer to:

https://github.com/robotology-playground/docker_images/blob/cb4ab232a327b03ae4c84fd9e3d2f0bb8a6fc213/robotology-tdd/Dockerfile#L71-L88

Bear in mind that this approach does not support ccache. Especially for big projects like this one, ccache support is a must-have.

traversaro:
My current plan is to actually add a CMake option to switch between *stable* (the old "master") and *unstable* (the old "devel") tags, documenting clearly that switching after the first project generation or having multiple builds with different settings is not supported at all. This is not ideal solution, but I think better then the status-quo. Once that is in place, I think it make sense to have GitHub Actions jobs dedicated to build them.

diegoferigo:
Even better :+1: I would add also the support to load a custom `ProjectTags*.cmake`.

traversaro:
Good idea.

nunoguedelha:
Let me know if you guys need any help with that.

PR #269, opened 2019-10-01
Set OsqpEigen_TAG also for devel braches

traversaro:
This should fix the build of superbuild with devel branches


traversaro:
To avoid this regressions, we should fix https://github.com/robotology/robotology-superbuild/issues/101 .

traversaro:
This should fix https://github.com/robotology/gazebo-yarp-plugins/issues/442 .

traversaro:
@gabrielenava I wonder if the reason behind https://github.com/robotology/robotology-superbuild/issues/263 was that the user was using the `devel` branches.

Issue #237, opened 2019-08-22
Fix compilation of ROBOLOGY_ENABLE_ICUB_HEAD  profile option on macOs


traversaro:
At the moment, the ROBOLOGY_ENABLE_ICUB_HEAD option fails on macOs with a dragonfly related error, see https://github.com/robotology/robotology-superbuild/pull/234/checks?check_run_id=200516475 .

traversaro:
Probably it is sufficient to add  https://brewformulas.org/Libdc1394 as a dependency.

traversaro:
cc @kouroshD did you experience any problem like that?

kouroshD:
No, it was fine for me. I did not have any problem.

traversaro:
cc @nunoguedelha

nunoguedelha:
@traversaro , so both the `superbuild` and `icub-main` Readmes should be updated right?

traversaro:
Exactly.

nunoguedelha:
I think we can also close this issue for now, considering https://github.com/robotology/robotology-superbuild/pull/268.

traversaro:
As the option is now tested in macOS as part of the CI with https://github.com/robotology/robotology-superbuild/pull/276, I agree that we can close the issue.

PR #274, opened 2019-10-15
Set dragonfly2 flag to OFF MacOS. Fix typos in ProjectTagsDevel.cmake

nunoguedelha:
Fixed the missing suffix `_TAG` in the tags defined in `ProjectTagsDevel.cmake` for `icub-models` and `icub-gazebo`.
Set `ENABLE_icubmod_dragonfly2` to OFF by default if building on a Mac. This option has for dependency the library **libdc1394** (Refer to http://www.icub.org/software_documentation/group__icub__dragonfly2.html and #237), which installation can be problematic on the MacOSX depending on the system version (Sierra, High-Sierra, Mojave, etc). => To be further analysed.


traversaro:
Thanks Nuno!

Issue #265, opened 2019-09-25
YCM build failing with Xcode


nunoguedelha:
## Issue description
The YCM build is failing when building the superbuild on Macos using Xcode toolchain.

Steps to reproduce the issue:
- update the `robotology-superbuild` repo to the latest commit on `master` branch today,
- run cmake on an empty `build` folder (`cmake ./ -G Xcode`),
- set to ON the options:
```
ROBOTOLOGY_ENABLE_CORE
ROBOTOLOGY_ENABLE_DYNAMICS
ROBOTOLOGY_ENABLE_ICUB_HEAD
ROBOTOLOGY_USES_GAZEBO
ROBOTOLOGY_USES_MATLAB
```
(This configuration has been used for a long time on the single motor test setup, for starting a set of EMS+2FOC boards driving a single iCub motor).
- a few repos are set with `YCM_EP_DEVEL_MODE` ON.
- But the following devel mode option is OFF:
```
YCM_EP_DEVEL_MODE_YCM
```
- run `xcodebuild -target ALL_UPDATE`,
- run `xcodebuild -target ALL_BUILD`.

The last commands fails.

## Failure log
```
...
PhaseScriptExecution CMake\ Rules build/robotology-superbuild.build/Debug/YCM.build/Script-5053C3E3F7E142C7BBCF2381.sh
    cd /Users/nunoguedelha/dev/robotology-superbuild
    /bin/sh -c /Users/nunoguedelha/dev/robotology-superbuild/build/robotology-superbuild.build/Debug/YCM.build/Script-5053C3E3F7E142C7BBCF2381.sh
echo "Skipping update step for 'YCM'"
Skipping update step for 'YCM'
cd /Users/nunoguedelha/dev/robotology-superbuild/robotology/YCM && /usr/local/Cellar/cmake/3.15.3/bin/cmake -E echo_append
echo "No patch step for 'YCM'"
No patch step for 'YCM'
/usr/local/Cellar/cmake/3.15.3/bin/cmake -E echo_append
/usr/local/Cellar/cmake/3.15.3/bin/cmake -E touch /Users/nunoguedelha/dev/robotology-superbuild/build/robotology/YCM/CMakeFiles/YCMStamp/Debug/YCM-patch
echo "Performing configure step for 'YCM'"
Performing configure step for 'YCM'
[...]

PR #276, opened 2019-10-19
Add macOS jobs to GitHub Actions

traversaro:


traversaro:
The problem reported in https://github.com/robotology/robotology-superbuild/issues/265 emerged, so this is blocked until we solve https://github.com/robotology/robotology-superbuild/issues/265 .

.github/workflows/ci.yml
```diff
@@ -18,7 +18,12 @@ jobs:
     strategy:
       matrix:
         build_type: [Debug, Release]
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, macOS-10.14]
+        include:
+          - os: ubuntu-latest
+            cmake_generator: "Unix Makefiles"
+          - os: macOS-10.14
+            cmake_generator: "Xcode"
```
diegoferigo: This is something of GHA that I haven't yet understood. What is the effect of the `include` entry? There is already a 2x2=4 matrix from the `build_type` and `os` tags, would the include add two more (as in Travis CI if I recall?) resulting into 6 jobs? It seems that only 4 start.

.github/workflows/ci.yml
```diff
@@ -43,12 +48,11 @@ jobs:
         sudo ./.ci/install_debian.sh
 
     - name: Dependencies [MacOS]
-      if: matrix.os == 'macOS-latest'
+      if: matrix.os == 'macOS-10.14'
       run: |
+        brew cask install xquartz
         brew install ace asio eigen tinyxml swig qt5 gsl pkg-config jpeg tinyxml
-        # Understand how to export env variables in a os-specific way
-        export Qt5_DIR=/usr/local/opt/qt5/lib/cmake/Qt5
```
diegoferigo: Curiosity: is this no longer necessary?

traversaro: I have no idea, but I wanted to test it.

.github/workflows/ci.yml
traversaro: As mentioned in the GitHub docs (in https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix): 
> You can add additional configuration options to a build matrix job that already exists. You cannot use include to add new jobs to a build matrix. For example, if you want to use a specific version of npm when the job that uses windows-2016 and version 4 of node runs, you can use include to specify that additional option.

Basically the include option does not add jobs, but just additional options to existing jobs.

diegoferigo: Ow this is cool, I was not aware of it. Then, job exclusion can be done with job-wise `if`s.

.github/workflows/ci.yml
```diff
@@ -43,12 +48,11 @@ jobs:
         sudo ./.ci/install_debian.sh
 
     - name: Dependencies [MacOS]
-      if: matrix.os == 'macOS-latest'
+      if: matrix.os == 'macOS-10.14'
       run: |
+        brew cask install xquartz
         brew install ace asio eigen tinyxml swig qt5 gsl pkg-config jpeg tinyxml
```
nunoguedelha: 'tinyxml' appears twice (actually also in the superbuild README).
What about: boost sqlite libedit ipopt opencv@3 ?

traversaro:
New problems emerged:
* https://github.com/robotology-playground/yarp-matlab-bindings/issues/42
* https://github.com/robotology/gazebo-yarp-plugins/issues/450

traversaro:
Rebased on the top of master to also test stable and unstable branches.

.github/workflows/ci.yml
traversaro: Apparently it is not necessary anymore, as the test are successful even without setting it.

traversaro: Apparently it is still necessary, as without it the Qt5 GUI are skipped, see in the log: 
~~~
2019-11-07T17:11:17.9116700Z -- Qt5 and GTK2 not found/selected, skipping iCubSkinGui
2019-11-07T17:11:17.9117690Z -- Qt5 and GTK2 not found/selected, skipping frameGrabberGui2
2019-11-07T17:11:17.9118630Z -- Qt5 and GTK2 not found/selected, skipping skinManagerGui
~~~

traversaro: @diegoferigo any idea on how to set `export Qt5_DIR=/usr/local/opt/qt5/lib/cmake/Qt5` only for a macos runner without duplicating the build step?

diegoferigo: Yes, there's an [undocumented method](https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps) that is all but future proof:

```yaml
- name: Setting env var
  if: matrix.os == macOS-10.14
  run: |
    echo "::set-env name=Qt5_DIR::/path/to/dir"
    env
```

traversaro: Ah cool. For the time being I just set `Qt5_DIR` to `/usr/local/opt/qt5/lib/cmake/Qt5` also in  Linux, as `/usr/local/opt/qt5/lib/cmake/Qt5` is an empty directory in Linux and qt is found just by using the system paths, but this method may be useful.

traversaro:
Besides some bindings problem that are tracked in https://github.com/robotology/robotology-superbuild/issues/290 and  https://github.com/robotology/robotology-superbuild/issues/291 , the PR is working fine, so let's merge it.

.github/workflows/ci.yml
traversaro: It seems that there are some docs about it, see : https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#set-an-environment-variable-set-env . Thanks @pattacini for the link.

diegoferigo: Cool, thanks for letting me know! Also the logger support is quite interesting

Issue #256, opened 2019-09-18
small mistake in README.m file (setup environment instructions on Linux system)


valegagge:
While I was installing the Superbuild at first time, I noticed a small mistake about setup environment instructions on Linux system.

In order to use the updated .bashrc (with `setup.sh`) it is necessary to add the definition of  `$ROBOTOLOGY_SUPERBUILD_INSTALL_PREFIX` variable before including `setup.sh`.

An other solution, is change the suggested command 
` source $ROBOTOLOGY_SUPERBUILD_INSTALL_PREFIX/share/robotology-superbuild/setup.sh `
in something like
` /directory/where/you/downloaded/robotology-superbuild/build/install `

Maybe this is only fussiness...

traversaro:
Indeed, I think it is a correct suggestion, feel free to suggest a PR.

traversaro:
Fixed  in https://github.com/robotology/robotology-superbuild/pull/278 .

valegagge:
Can we close the issue?

traversaro:
Yes, thanks!

PR #278, opened 2019-10-21
Fix setup environment instructions on Linux system

valegagge:
This PR should fix issue #256 

@traversaro


traversaro:
Thanks @valegagge !

PR #288, opened 2019-11-06
Add ICUBcontrib to walking-controllers dependencies

traversaro:
In https://github.com/robotology/walking-controllers/pull/44 the `ICUBcontrib` CMake package was added to the walking-controllers dependencies, but we never updated the related file in the superbuild.


traversaro:
Thanks @GiulioRomualdi , let's wait for CI to merge.

PR #291, opened 2019-11-08
DO NOT MERGE - Test commit to reproduce gazebo problem

traversaro:
See https://github.com/robotology/gazebo-yarp-plugins/issues/447


traversaro:
We were not able to reproduce the issue discussed in https://github.com/robotology/robotology-superbuild/pull/291 .

PR #296, opened 2019-11-12
Add Windows support in GitHub Actions

traversaro:


traversaro:
The CI is not complete (the disabled part are documented in the scripts with the reference to the relative issue) but not it runs fine, so I think we can merge it as it is and modify it as soon as the blocking issue get resolved.

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       matrix:
         build_type: [Debug, Release]
         project_tags: [Stable, Unstable]
-        os: [ubuntu-latest, macOS-10.14]
+        os: [ubuntu-latest, macOS-10.14, windows-2019]
```
diegoferigo: I didn't find this key in the official documentation. Where did you find it?

traversaro: Apparently both windows-2019 and macOS-10.14 disappeared from the documentation, but they used to be there and they are still working fine. See https://github.com/actions/virtual-environments/issues/81 for a related issue for macOS .

diegoferigo: I see, thanks for the link!

PR #300, opened 2019-11-18
Added USB camera yarp device in icub-head profile

valegagge:
With this PR the `usbCamera yarp device` is compiled if `icub-head profile` is active.
Since some robots mount usb cameras, it is important to make the corresponding device available.

To achieve this, I changed the cmake file ```cmake/BuildYARP.cmake ``` enabling the usbCamera cmake option only on Linux system.
Moreover I updated the installation instructions for Linux system by adding `libv4l-dev` library, because `usbCamera`is based on `video4linux library`.

I verified the `usbCamera` device is available using `yarpdev --list` command. I think this check besides CI tests, could be sufficient to verify that these changes could be ok.

### DIFF
--- a/.ci/install_debian.sh
+++ b/.ci/install_debian.sh
@@ -10,7 +10,7 @@ export DEBIAN_FRONTEND=noninteractive
 apt-get install -y clang valgrind ccache ninja-build
 
 # Core dependencies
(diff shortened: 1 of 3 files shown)


traversaro:
Can you also add the `libv4l-dev` dependency in https://github.com/robotology/robotology-superbuild/blob/master/.ci/install_debian.sh#L13, so it is actually tested by the CI ? Thanks!

valegagge:
> Can you also add the `libv4l-dev` dependency in https://github.com/robotology/robotology-superbuild/blob/master/.ci/install_debian.sh#L13, so it is actually tested by the CI ? Thanks!

yes, of course! Done [here](https://github.com/robotology/robotology-superbuild/pull/300/commits/8af1e100b8189d23ca77f89a5dfb5da74e6ed4cf) :)

Issue #213, opened 2019-06-20
Add a testing profile


Nicogene:
Is there a way to compile the `robotology-superbuild` skipping the test part? (`robottestingframwork` and `icub-tests`)

If not, it would be nice adding it, since in some cases it is not useful and can give some compilation troubles (see robotology/yarp#2050).

traversaro:
I am a bit afraid about adding more cmake options, beside the profile ( https://github.com/robotology/robotology-superbuild#profile-cmake-options ) and dependencies one ( https://github.com/robotology/robotology-superbuild#dependencies-cmake-options ), as the project configuration is already quite complex as it is to maintain and understand for users. Something that I think could make sense is to add a "testing" profile that includes rtf, icub-tests and eventually blocktest, so that it would be possible to disable those projects via the usual profile systems. Do you think it would make sense?

Nicogene:
Fine for me :+1:

traversaro:
Ok for me. For anyone interested, please open a PR with this change.

PR #303, opened 2019-11-24
Create new Robot Testing profile for compilation  of rtf and icub-tests

traversaro:
See https://github.com/robotology/robotology-superbuild/issues/213 for an extensive description of the change. Note that the default for this profile is `OFF`, so this is a slight change w.r.t. to the past where robot-testing-framework and iCub Tests were compiled by default, but I think it make sense to do so. 

Fix https://github.com/robotology/robotology-superbuild/issues/213 .


traversaro:
cc @valegagge @vtikha this slim downs the amount of software necessary to compile for the core profile of the superbuild.

Issue #128, opened 2018-10-20
Get Windows dependencies from NuGet/Chocolatey repository 


traversaro:
As part of the ROS on Windows effort (http://aka.ms/ros ) Microsoft is maintaining a NuGet/Chocolatey repository of C++ libraries in binary form: https://roswin.azurewebsites.net/packages . I never considered Chocolatey as a serious option to obtain the dependencies on Windows because the number of C++ libraries available in their main repo is quite limited, but this new repo contains most of our dependencies. I am opening this issue to track the possibilities of using this packages to build the robotology-superbuild on Windows. 

Related ROS Discourse discussion: https://discourse.ros.org/t/moveit-now-enabled-for-ros1-on-windows/6509/2 .

Packages currently required by the robotology-superbuild but not available on the ROS NuGet repo: 
* ACE 
* GSL
* IPOPT 
* ODE
* libxml2 

cc @drdanz @claudiofantacci @pattacini

traversaro:
Another interesting option that emerged only recently is https://bincrafters.github.io/ .

traversaro:
I had to reinstall the superbuild on Windows so I gave it a try. As mentioned in https://discourse.ros.org/t/moveit-now-enabled-for-ros1-on-windows/6509/2 , trying to build in Debug fails as RTF uses TinyXML that in its API uses STL objects (cc @claudiofantacci with which we discussed about it): 
~~~
Severity	Code	Description	Project	File	Line	Suppression State
Error	LNK1319	6 mismatches detected	testrunner	C:\src\robotology-superbuild\build-nuget\robotology\RTF\bin\Debug\testrunner.exe	1	
Error	LNK2038	mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in ErrorLogger.obj	testrunner	C:\src\robotology-superbuild\build-nuget\robotology\RTF\src\testrunner\tinyxml.lib(tinyxml.obj)	1	
Error	LNK2038	mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug' in ErrorLogger.obj	testrunner	C:\src\robotology-superbuild\build-nuget\robotology\RTF\src\testrunner\tinyxml.lib(tinyxml.obj)	1	
[...]

PR #304, opened 2019-11-25
Add CI build using  binaries provided  by ROS's Chocolatey

traversaro:
Related  issue: https://github.com/robotology/robotology-superbuild/issues/128 . 
Thanks to the integration of vcpkg in ROSOnWindows infrastructure, we can just use Chocolatey packages for most dependencies, and install the missing ones using vcpkg. 
The major advantages is that in this way we can start enabling in CI the compilation of the ROBOTOLOGY_USES_GAZEBO option on  Windows.


traversaro:
Not working for an ipopt problem, closing for now.

PR #308, opened 2019-11-28
Update installations docs to mention that CMake 3.12 is required and require it at CMake level

traversaro:
Following the announcement in https://github.com/robotology/QA/issues/364 and the discussion  in  https://github.com/robotology/ycm/issues/261 ,  we require CMake 3.12 now . This PR updates the documentation, indicating the recommended way to install a recent CMake in Debian 9 and Ubuntu 18.04 .


traversaro:
cc @valegagge @vtikha this probably means that the docker scripts need to be updated.

PR #316, opened 2019-12-09
User install prefix

valegagge:
In this PR I used the YCM variable called ```YCM_EP_INSTALL_DIR``` in order to let user to choose the installation prefix.

Please see [here](https://github.com/robotology/ycm/pull/289) for more details about ```YCM_EP_INSTALL_DIR```

I tested these changes on Linux platform only.

Thanks to @drdanz 

cc @traversaro


CMakeLists.txt
```diff
@@ -219,7 +219,7 @@ endif()
 
 include(ConfigureFileWithCMakeIf)
 if(NOT WIN32)
-  configure_file_with_cmakeif(${CMAKE_CURRENT_SOURCE_DIR}/cmake/template/setup.sh.in ${CMAKE_BINARY_DIR}/install/share/${PROJECT_NAME}/setup.sh @ONLY)
+   configure_file_with_cmakeif(${CMAKE_CURRENT_SOURCE_DIR}/cmake/template/setup.sh.in ${CMAKE_BINARY_DIR}/install/share/${PROJECT_NAME}/setup.sh @ONLY)
```
traversaro:
```suggestion
  configure_file_with_cmakeif(${CMAKE_CURRENT_SOURCE_DIR}/cmake/template/setup.sh.in ${CMAKE_BINARY_DIR}/install/share/${PROJECT_NAME}/setup.sh @ONLY)
``
Sorry for the nitpick, but let's try to keep the 2-space indentation

traversaro:
I also noticed that a similar fix should be also done in https://github.com/robotology/robotology-superbuild/blob/89279f3f4c263819893bdc52c259ce39e7de43e4/cmake/YCMBootstrap.cmake#L72 , but that files comes from YCM , see https://github.com/robotology/ycm/blob/7cc2b755e302b972112ca6b77a6dbce4492d0062/tools/YCMBootstrap.cmake#L75 . I guess that it is not a big problem as  `YCM_EP_INSTALL_DIR` explicitly does not support to build for a second-time a project, but it may be worth to check it @drdanz .

valegagge:
I fixed all template files in e12d969afa1e8401fa4198a6cae09834f32a9116.

While, about 

> I also noticed that a similar fix should be also done in
> 
> https://github.com/robotology/robotology-superbuild/blob/89279f3f4c263819893bdc52c259ce39e7de43e4/cmake/YCMBootstrap.cmake#L72
> 
> , but that files comes from YCM , see https://github.com/robotology/ycm/blob/7cc2b755e302b972112ca6b77a6dbce4492d0062/tools/YCMBootstrap.cmake#L75 . I guess that it is not a big problem as `YCM_EP_INSTALL_DIR` explicitly does not support to build for a second-time a project, but it may be worth to check it @drdanz .

I''ll check it with @drdanz .

drdanz:
This is a bit tricky, because that check removes from the path the install dir, so that after the bootstrap the find package will not find the bootstrapped version at the second run. On the other hand, as documented, the `YCM_EP_INSTALL_DIR` is supposed to be used only for building the superbuild *once*.

Removing (for example) `/usr/local` from `CMAKE_PREFIX_PATH` and `PATH` would just be wrong, because it will cause problems in finding other packages...

I'd say not to change that file, but I'd like to hear also @traversaro opinion...

traversaro:
> 
> 
> This is a bit tricky, because that check removes from the path the install dir, so that after the bootstrap the find package will not find the bootstrapped version at the second run. On the other hand, as documented, the `YCM_EP_INSTALL_DIR` is supposed to be used only for building the superbuild _once_.
> 
> Removing (for example) `/usr/local` from `CMAKE_PREFIX_PATH` and `PATH` would just be wrong, because it will cause problems in finding other packages...
> 
> I'd say not to change that file, but I'd like to hear also @traversaro opinion...

This is a really good point, and a good reason for why `YCM_EP_INSTALL_DIR` can only be supported for installing the superbuild once.

traversaro:
The CI failure are unrelated, and should be fixed by https://github.com/robotology/robotology-superbuild/pull/317 .

traversaro:
cc @PeterBowman that wanted to experiment with `YCM_EP_INSTALL_DIR` .

traversaro:
> I fixed all template files in [e12d969](https://github.com/robotology/robotology-superbuild/commit/e12d969afa1e8401fa4198a6cae09834f32a9116).

Ok, if you accept the suggestion in https://github.com/robotology/robotology-superbuild/pull/316#discussion_r355532260 , the PR for me is ready to be merged.

valegagge:
> > I fixed all template files in [e12d969](https://github.com/robotology/robotology-superbuild/commit/e12d969afa1e8401fa4198a6cae09834f32a9116).
> 
> Ok, if you accept the suggestion in [#316 (comment)](https://github.com/robotology/robotology-superbuild/pull/316#discussion_r355532260) , the PR for me is ready to be merged.

Done! thanks!

traversaro:
Thanks! I do not think that Windows test are necessary, if you already tested the template generation logic on Linux.

PR #320, opened 2019-12-12
Switch stable branch of YARP from 3.2 to 3.3

traversaro:
See https://github.com/robotology/QA/issues/371 for more details.

~~**DO NOT MERGE** (i.e. the CI will fail) until all the v1.14 iCub's repo have been released, as I am afraid that the v1.13 iCub repos are not compatible with  YARP 3.3.0 .~~  

It turns out the v1.13 iCub repos are compatible with YARP 3.3, so we can merge without problems even before the v1.14 iCub releases.


traversaro:
It turns out the v1.13 iCub repos are compatible with YARP 3.3, so we can merge without problems even before the v1.14 iCub releases.

Issue #242, opened 2019-09-01
Use latest tagged version of upstream osqp and osqp-eigen instead of local fork


traversaro:
Apparently the robotology-superbuild is currently using a fork of osqp for dealing with some `.gitattributes` problems (see https://github.com/robotology/robotology-superbuild/pull/87/files#diff-fcad91dd1d4b0f9e2cc5f258bfa8b79bR10). As new version of osqp are being released, and sometimes they also contain breaking changes (see https://github.com/robotology/osqp-eigen/issues/31) we should switch to use upstream osqp, optionally on a specific release tag, so that we can update to use new osqp version by just changing the tag.

traversaro:
The original issue was discussed in https://github.com/robotology/robotology-superbuild/pull/87#issuecomment-458987172 and the following comments. Dropping Ubuntu 16.04 should be sufficient to solve the problem.

traversaro:
As of  https://github.com/robotology/robotology-superbuild/pull/243, the issue seem to be solved, probably due to using the latest ExternalProject module via requiring the latest YCM (see https://github.com/robotology/robotology-superbuild/pull/224 and https://github.com/robotology/robotology-superbuild/pull/199).

traversaro:
In https://github.com/robotology/robotology-superbuild/pull/260 we used a custom tag for osqp-eigen with our fork of osqp. To fix this we need to update all the software that depends on osqp-eigen in the superbuild to deal with this new version, and then we can update. 

@GiulioRomualdi let me know if you need my help on this.

GiulioRomualdi:
@S-Dafarra check this: https://github.com/robotology/robotology-superbuild/issues/242#issuecomment-526896522 :smile:

traversaro:
Just for clarity, I think the incompatibility for which the fork was necessary was just for Ubuntu 16.04 + GitHub Actions, not Ubuntu 16.04 on its own (that in any case we stopped testing in CI).

PR #330, opened 2020-01-17
Update osqp to v0.6.0

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/242 .


traversaro:
@GiulioRomualdi apparently OsqpEigen `devel` is not compatible with osqp 0.6.0 .  Do you think it  make then sense to use `master` as the target branch for both "Stable" and  "Unstable" builds.

GiulioRomualdi:
> @GiulioRomualdi apparently OsqpEigen `devel` is not compatible with osqp 0.6.0 . Do you think it make then sense to use `master` as the target branch for both "Stable" and "Unstable" builds?

That's really strange, `devel` is on top of `master`

GiulioRomualdi:
CI is failing because of https://github.com/robotology/osqp-eigen/blob/master/CMakeLists.txt#L105

Similar to: https://github.com/robotology/idyntree/pull/629

traversaro:
> 
> 
> CI is failing because of https://github.com/robotology/osqp-eigen/blob/master/CMakeLists.txt#L105
> 
> Similar to: [robotology/idyntree#629](https://github.com/robotology/idyntree/pull/629)

Yes, sorry I overlooked the error message. How do you prefer to proceed? In general it may be expected that not all the unstable branches of all the project of the robotology-superbuild, are compatible with one another, so in the "Unstable" branches we actually just collect an arbitrary compatible set of branches, but there is no strict requirement to have the unstable branches of all dependencies. If you prefer we can just use the `master` branch of `osqp-eigen` with the `devel` branch of `walking-controllers` .

GiulioRomualdi:
https://github.com/robotology/osqp-eigen/pull/42 should fix the problem

GiulioRomualdi:
`SameMinorVersion` was a terrible workaround to https://github.com/oxfordcontrol/osqp/issues/195.
`osqp-eigen v0.5.1` is back-compatible with `v0.4.1` so there shouldn't be any problems

traversaro:
Thanks @GiulioRomualdi . I would wait for next week to merge this, to hear if someone needs to stick to osqp 0.5 .

PR #337, opened 2020-01-29
Configure setup scripts directly in YCM_EP_INSTALL_DIR instead of hardcoding ${CMAKE_BINARY_DIR}/install

traversaro:
This may require that `cmake` is executed as `sudo` if `YCM_EP_INSTALL_DIR` points to `/usr/local`, but in any case the use of `YCM_EP_INSTALL_DIR` is an experimental feature reserved only to advanced setups, so I do not think it is a big problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,13 +219,13 @@ endif()
 
 include(ConfigureFileWithCMakeIf)
 if(NOT WIN32)
-  configure_file_with_cmakeif(${CMAKE_CURRENT_SOURCE_DIR}/cmake/template/setup.sh.in ${CMAKE_BINARY_DIR}/install/share/${PROJECT_NAME}/setup.sh @ONLY)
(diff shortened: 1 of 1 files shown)


traversaro:
cc  @valegagge

traversaro:
@valegagge I would like to merge this PR, but this is probably going to be break your docker  image generation scripts, I do not know how do you prefer to proceed.

valegagge:
@traversaro 

> @valegagge I would like to merge this PR, but this is probably going to be break your docker image generation scripts, I do not know how do you prefer to proceed.

Feel free to merge this pull request, I'll update our image in according with these changes.
Thank you! :)

cc @vtikha

traversaro:
cc @mbrunettini I just realized that without this change the Windows installer will do not set the correct env variables, so you may want to update the superbuild to the latest master  now that this PR has been merged.

PR #341, opened 2020-02-01
Test DYNAMICS profile under Windows

traversaro:
Remove workaround for https://github.com/robotology/blockfactory/pull/45

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,9 +123,9 @@ jobs:
       shell: bash
       run: |
         mkdir -p build
-        cd build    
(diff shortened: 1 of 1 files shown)


traversaro:
Now the issue fails due to https://github.com/robotology/walking-controllers/issues/47 , that is a good example that we would have catched before if the CI was working. :D:

traversaro:
> 
> 
> Now the issue fails due to [robotology/walking-controllers#47](https://github.com/robotology/walking-controllers/issues/47) , that is a good example that we would have catched before if the CI was working. :D:

Fix for that issue provided in https://github.com/robotology/walking-controllers/pull/53 .

Issue #333, opened 2020-01-28
Request for Comment: move all source repos in src subdirectory 


traversaro:
I would like to collect feedback on the following: instead of downloading the source repo of subprojects in the `external` or `robotology` directories as it is now, download all the source repos in the `src` directory. 

Pros: 
* Consistency: all the source repos will go in the same directory, and with a directory whose meaning is clear from its name. 
* Familiarity for people used to ROS/ROS2/colcon: in ROS/ROS2 workspaces, all the repos are contained in the `src` subdirectory of the  workspace 
* We can use more easily the `vcs` tool (see https://github.com/dirk-thomas/vcstool)  to manipulate the repos if they are in devel model, for example to export the value of all the repos in the superbuild. `vcs` is the tool used in ROS2, so it would be something familiar used to ROS2 workspaces. In theory `vcs` could be used also with the current directory structure, but the saved file would have the strange format of having the `external` or `robotology` directory prefixed to the CMake package name: 
~~~
repositories:
  external/osqp:
    type: git
    url: https://github.com/oxfordcontrol/osqp
    version: master
  robotology/YARP:
    type: git
    url: https://github.com/robotology/yarp
    version: master
~~~
this would complexify its use as a reference for superbuild tags, while using a single `src` directory would result in a much more natural: 
~~~
repositories:
  osqp:
    type: git
    url: https://github.com/oxfordcontrol/osqp
    version: master
  YARP:
    type: git
    url: https://github.com/robotology/yarp
    version: master
~~~

Cons:
* It is a change. : )

S-Dafarra:
I have no objections.

GiulioRomualdi:
It's fine!

valegagge:
For team-code is ok!
@vtikha

S-Dafarra:
In the meanwhile, I wrote a simple script to check the status of each repo:
```sh
#!/bin/bash

subdirs="$@"

if (($# == 0)); then
    subdirs="."
fi

for subdir in ${subdirs}; do \
    for i in ${subdir}/*/; do \
        echo "--------------------------------------------"
        (cd $i && echo -n "${i}: " \
        && git rev-parse --abbrev-ref HEAD \
        && git log -1 --format="%cr|%s" \
        && git status --short); \
    done
done
echo "--------------------------------------------"
```
It can be called with no arguments, or with the list of directories to check. I don't know if it is useful or if this is the right place.

traversaro:
[...]

PR #343, opened 2020-02-08
Add ycm_load_vcs_yaml_info CMake function

traversaro:
This function permits to load the `<package>_TAG` and `<package>_REPOSITORY` variables from a YAML file.

In particular, the YAML format used is the one produced and consumed by the [`vcstool`](https://github.com/dirk-thomas/vcstool) command line tool, that is used in ROS2 tutorials to manage groups of repositories. 

For example, the existing  file `ProjectsTagsStable.cmake` (see https://github.com/robotology/robotology-superbuild/blob/34325b12084672d1869695366ca53bb4ffe83a7f/cmake/ProjectsTagsStable.cmake) file could be refactored in:
~~~cmake
include(YCMLoadVcsYamlInfo)

ycm_load_vcs_yaml_info(YAML_FILE project_tags_stable.repos)
~~~ 

With `project_tags_stable.repos` being: 
~~~yaml
repositories:
  osqp:
    type: git
    url: https://github.com/oxfordcontrol/osqp.git
    version: v0.6.0
  YARP:
    type: git
    url: https://github.com/robotology/yarp.git
    version: yarp-3.3
~~~

The main advantages of this are two: 
* yaml file can be easily parsed also by non-CMake based software @vtikha @valegagge 
* once all the source repos are contained in the same directory (https://github.com/robotology/robotology-superbuild/issues/333) yaml files  corresponding to a precise set of commits can be easily created with `vcs export --exact > my.repos` @CarlottaSartore 

The script currently supports just a limited subset of the possible repos supported by YCM superbuilds, but once we test it a bit it could then make sense to move it in YCM @drdanz

### DIFF
--- a/cmake/YCMLoadVcsYamlInfo.cmake
+++ b/cmake/YCMLoadVcsYamlInfo.cmake
@@ -1,0 +1,108 @@
+#.rst:
+# YCMLoadVcsYamlInfo
+# -------
+#
+# Load YCM project variables for a YCM superbuild from a VCS yaml file.
+#
+# .. command :: ycm_load_vcs_yaml_info
+#
+#  Extract information about one commit from one git repository in
+#  ``SOURCE DIR``::
+#
+#    ycm_load_vcs_yaml_info(YAML_FILE <yaml_file>]
+#                           <VERBOSE>)
+#
+#  ``YAML_FILE`` should point to a yaml file in the format supported by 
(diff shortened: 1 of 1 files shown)


traversaro:
> Beyond feeling the pain of parsing files in plain CMake, it looks good to me :)

gottado

Issue #344, opened 2020-02-11
Compilation of Unstable branches with ROBOTOLOGY_USES_OCTAVE and ROBOTOLOGY_USES_MATLAB is broken


traversaro:
Due to yarp-matlab-bindings and https://github.com/robotology/yarp/issues/2200 .

traversaro:
Now blocked by https://github.com/robotology/yarp-matlab-bindings/issues/44 .

traversaro:
Fixed by: 
* https://github.com/robotology/robotology-superbuild/pull/345
* https://github.com/robotology/yarp-matlab-bindings/pull/46

traversaro:
Both the aforementioned PRs have been merged.

PR #346, opened 2020-02-16
Disable fail-fast in GitHub Actions

traversaro:
Due to the nature of robotology-superbuild, often the unstable branches build fails. However, it is important that even if the unstable branches build fails the stable branches build continue to build, so we are still able to detect regressions on the stable branches, and  just merge PRs if the stable branches pass.


traversaro:
The unstable branches will probably fail due to https://github.com/robotology/robotology-superbuild/issues/344, but if the stable branches pass I think we can merge the PR.

Issue #227, opened 2019-07-11
Add Release versions for superbuild


kouroshD:
When there are new updates on the core repositories of the superbuild, e.g. YARP, RobotTestingFramework, ICUB, etc , normally there are unseen bugs to fix (build/run time problems). Also, other repositories are based on the mentioned ones, and therefore they might be affected.
I am suggesting here to add releases for superbuild, so that when there are updates like recent updates on YARP, RTF, etc. , we can go to a release of the superbuild, and we can install the packages with Tags (maybe) different from `HEAD`. 
For example, now HDE works with the YARP version `v3.1.1` not with the new version, but then this version is not compatible with new versiona of RTF, ... .
This suggestion may increase the time for fixing the bugs of new released repositories, but on the other side, it does not stop others from developments.
cc @DanielePucci

diegoferigo:
The support of defining releases has been introduced in https://github.com/robotology/robotology-superbuild/pull/100. Users can create a new tag file (potentially in a new branch of this project if it has to be shared with other users) waiting all upstream post-release fixes.

The current status unfortunately is a corner case since we changed few things in the superbuild structure (RTF as you wrote).

I can think of a per-profile tags (e.g. andy), but handling release constraints of multiple profiles that might be conflicting is not easy.

kouroshD:
Making a new branch will work as you mention, but defining releases in a new project may not always work, e.g., our current situation and superbuild should be in previous commits. Well, I am not sure if having multiple branches is a good response to this problem.

The profiles defined in superbuild (especially `core` and `dynamics`) they should be compatible to each other, and they should provide functionality the user is seeking.

diegoferigo:
Having multiple branches is not the right solution, but it might address the corner case we encountered this case (that is the impossibility to roll back to old releases by just changing the project TAG). Generally, it should not be required.

> I am suggesting here to add releases for superbuild, so that when there are updates like recent updates on YARP, RTF, etc. , we can go to a release of the superbuild, and we can install the packages with Tags (maybe) different from HEAD.

What you wrote in the first post (use superbuild releases) would not solve it because for most projects we do not check-out specific commits. This is clear in the [devel case](https://github.com/robotology/robotology-superbuild/blob/master/cmake/ProjectsTagsDevel.cmake). We point to branches, not commits.

Using releases would mean checking out the repository to a specific point in history. Referring to the RTF change we introduced, rolling back would still checkout the `master` branch of both yarp and rtf.

[...]

PR #350, opened 2020-02-20
Hardcode the ROBOTOLOGY_PROJECT_TAGS_CUSTOM_FILE to 2020.02.yaml for the releases/2020.02 branch

traversaro:
This commits hardcodes the ROBOTOLOGY_PROJECT_TAGS_CUSTOM_FILE to point to the  2020.02.yaml file for the `releases/2020.02` branch. The idea  is to tag the `v2020.02` tag on this branch (and eventually other patch releases of the v2020.02 series), to ensure that a user of the superbuild just needs to checkout the correct tag of the superbuild, and then the superbuild will download the correct versions of the subprojects. 

Note that is quite important to **never forward merge** the `releases/**` branches to `master`.


traversaro:
I think this finally fixes https://github.com/robotology/robotology-superbuild/issues/227, what do you think @kouroshD ?

kouroshD:
> I think this finally fixes #227, what do you think @kouroshD ?

 Definitely it fixes that issue.

PR #351, opened 2020-02-20
Use version 0.0.3 of the pre-compiled vcpkg workspace in CI

traversaro:
This is required to be compatible with the installer scripts provided in https://github.com/robotology/robotology-superbuild/pull/334 .
Other related issue: https://github.com/robotology-playground/robotology-superbuild-dependencies-vcpkg/pull/5 .


traversaro:
As the change affects only Windows, we can ignore the fact that macOS builds still needs to be finished and merge them anyway.

Issue #324, opened 2020-01-03
Support OpenCV 4  


traversaro:
We documented the use of opencv 2 or 3 in https://github.com/robotology/robotology-superbuild/issues/169 , but given that OpenCV 4 will be used in Ubuntu 20.04 , we need to make sure that all profiles of the superbuild work fine with OpenCV 4 . @Nicogene did some good work in solving the issue for YARP at https://github.com/robotology/yarp/pull/2162 , once that is merged we should test the superbuild and see if it is able to work correctly with OpenCV 4 , or if other projects (such as icub-main) need to be fixed as well.

traversaro:
Downstream issues: 
* yarp: 
  * https://github.com/robotology/yarp/issues/1672
  * https://github.com/robotology/yarp/pull/2162
* icub-main : 
  * https://github.com/robotology/icub-main/issues/634

traversaro:
I tested in https://github.com/robotology/robotology-superbuild/pull/327 compilation against OpenCV 4, but there are still compilation issues in icub-main : https://github.com/robotology/icub-main/issues/634 .

traversaro:
As far as I know, all the projects in the robotology-superbuild has been updated to support OpenCV, at least in their unstable branches. We do not switch to opencv4 in the CI of the robotology-superbuild to avoid breaking the stable branches build, but I think that we can  close this issue.

PR #357, opened 2020-02-25
Use default opencv in Homebrew on macOS

traversaro:
Updated both the GitHub Action and the README.


traversaro:
We can do that thanks to all the hard work that was  done in all repositories, see https://github.com/robotology/robotology-superbuild/issues/324 and the linked issues.

PR #359, opened 2020-02-29
Automatically recognize the active Python version

diegoferigo:
The superbuild currently defaults to Python 2.7, even though the majority of our projects either deprecated this version or default to 3.6.

I tested this change only on Ubuntu 18.04. I am not sure why my superbuild creates `site-packages` instead of `dist-packages`. Can we check on other systems what is the situation? I don't think there are many Python users of the superbuild, and this path problem might have been not noticed until now.

Note that `FindPython` finds python3 by default, searching for python2 only if python3 is not available.


diegoferigo:
Adding more information about the new `FindPython` module, I found the following variables:

- `Python_EXECUTABLE=/rl/ve/bin/python3.6`
- `Python_STDLIB=/usr/lib/python3.6`
- `Python_STDARCH=/usr/lib/python3.6`
- `Python_SITELIB=/rl/ve/lib/python3.6/site-packages`
- `Python_SITEARCH=/rl/ve/lib/python3.6/site-packages`

Note that in `/rl/ve` there's my the active virtualenv. Considering the purpose of the superbuild these variables are useless to us since the files are always installed in the `build/install` folder.

Though, if downstream projects use those variables to configure the install path and the superbuild does not override them, we might have some leaked files outside the `build/install` folder (e.g. inside the virtualenv or the current SITEARCH / SITELIB directory). This could happen because those `Python_*` variables seem to be absolute, and not relative to the install prefix.

traversaro:
> I am not sure why my superbuild creates `site-packages` instead of `dist-packages`. Can we check on other systems what is the situation?

I think this was just based on the iDynTree and YARP bindings Python bindings buildsystems, that just installed in the directory returned by the following python script (see https://github.com/robotology/yarp/blob/v3.3.2/bindings/python/CMakeLists.txt#L47 and https://github.com/robotology/idyntree/blob/master/bindings/python/CMakeLists.txt#L27): 
~~~
from distutils import sysconfig; print(sysconfig.get_python_lib(1,0,prefix=''))
~~~
Back in time in some configuration this script returned a prefix that contained `dist-packages`  and then I hardcoded that. Given that if I remember correctly also @drdanz was not really conviced by this installation logic, if you have better suggestions we can eventually change where YARP and iDynTree install the python binaries. if we instead stick with using `from distutils import sysconfig; print(sysconfig.get_python_lib(1,0,prefix=''))`, probably it make sense to just execute the same script also in the superbuild for consistency.

traversaro:
The stable failures are actually due to https://github.com/robotology/yarp/issues/2217 , sorry for this.

traversaro:
> The superbuild currently defaults to Python 2.7, even though the majority of our projects either deprecated this version or default to 3.6.

Just to clarify, YARP and iDynTree now for example in Ubuntu default to Python 3? In any case, I totally for just supporting Python 3 in the superbuild, so if necessary we can add the necessary CMake options to force iDynTree and YARP to compile their bindings for Python 3.

diegoferigo:
> if you have better suggestions we can eventually change where YARP and iDynTree install the python binaries

In my opinion is a complicated situation for the following reasons:

- The superbuild is self contained, it should install all its files inside `build/install` and the system should remain clean
- If the superbuild installs binary files (e.g. bindings) in the system, they will likely go inside a virtualenv tree (or, as backup, the user's home folder). Installing files from CMake inside this tree is somehow a dirty solution, since those files won't be tracked by pip.
- I would tend to prefer sticking with the current approach, that means installing python binaries in `build/install/lib/python3.6/site-packages` and adding this folder manually to the `PYTHON_PATH` (or, better, use the scripts).

However, this approach is not very pythonic. In fact, there's no pip involved. An alternative solution to provide python packages would be adding to all the subprojects of the superbuild (yarp, idyntree, icub-main) a `setup.py` file. In this way, python packages could be created.

Though, also with the pip way dependencies are tricky to handle. For example, icub-main will depend on yarp. So the yarp python package should contain all the shared libraries, headers, and cmake imported targets. I suspect that also in this case the user has to define at least a new entry of the `CMAKE_INSTALL_PREFIX` with the virtualenv folder. 

Anyway, this pip way goes out of the scope of this PR and it could be discussed separately somewhere else.

> I think this was just based on the iDynTree and YARP bindings Python bindings buildsystems, that just installed in the directory returned by the following python script (see https://github.com/robotology/yarp/blob/v3.3.2/bindings/python/CMakeLists.txt#L47 and https://github.com/robotology/idyntree/blob/master/bindings/python/CMakeLists.txt#L27)

Note that the command you reported is also used in the new `FindPython`, though in that case an *absolute* path is found. With the lines you posted, instead, the prefix is stripped. (e.g. `/rl/ve/lib/python3.6/site-packages` -> `lib/python3.6/site-packages`). Note also that if you have a virtualenv enabled, you would get `site-packages`, instead if you're using the system interpreter, `dist-packages` is returned.

> Just to clarify, YARP and iDynTree now for example in Ubuntu default to Python 3? 

I'd like to try on a clean system with a clean build folder, but I think yes. Python3 is recognized and used by default. I'm pretty sure about iDynTree, I didn't check YARP.

---

`TL;DR`

Keeping the current superbuild approach, I would double-check that all the subprojects install Python bindings in their own install prefix (e.g. `<prefix>/lib/python3.6/site-packages`), and require adding the same folder to the `PYTHON_PATH` env var.

If users want to install the projects with pip in a more pythonic way, the superbuild is not required. All those project must contain a `setup.py`, properly handing dependencies and installing also development files such shared libraries and headers.

traversaro:
Ok, so do you think we can merge this PR as it is, or do you prefer to wait and do some  more modifications?

> Anyway, this pip way goes out of the scope of this PR and it could be discussed separately somewhere else.

Yes, I think that expecting the superbuild to install in pip install prefix is fundamentally ill-posed. `pip` is a package manager, and as we don't expect the superbuild to install the packages in the prefix used by the package manager for `apt`, `homebrew` or `vcpkg`, I don't think we should expect something different for `pip`. In a sense, the superbuild is in itself a sort of a mini package manager on its own.

Leaving aside the superbuild, I think the well posed question is: "how we could support easy to use pip packages for yarp and idyntree? This is a well posed question, for which I personally don't have an answer. In  particular, from my limited experience pip does not provide any kind of facility for handling a complex tree of C++  dependencies, so basically we either assume that the C++ libraries are already installed in the system (as we do with ignition libraries) or we basically embed the build logic of the YARP|iDynTree library **and its dependencies** in the python pip build script (but then it is impossible to have a C++ library shared among two different C++-based pip packages. 
Interestingly, conda (with the conda-force ports) and Julia Pkg (with [BinaryBuilder](https://binarybuilder.org/) + [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil) have the infrastructure  to deal with projects with a complex tree of C++ dependencies. In particular the binary provided by Julia's BinaryBuilder seems not to be Julia specific at all, so perhaps they could be exploited in pip packages to deal with C++ dependencies. Unfortunately, the BinaryBuilder  does not support as of now the MSVC ABI, that is the one used by Python 3.6 on Windows, so this would not work on Windows.

diegoferigo:
> Ok, so do you think we can merge this PR as it is, or do you prefer to wait and do some more modifications?

Yes. The PR should also be robust in the case the default interpreter is Python 3.7, even though I didn't test such setup.

> I think the well posed question is: "how we could support easy to use pip packages for yarp and idyntree?

As you wrote, handling C++ dependencies with pip is a pain. An we also have to separate the two cases of users that install sdist and bdist.

In the **bdist case**, if all projects have binary packages in PyPI statically compiled, there should be no problem (watch out about plugin libraries!). Packages like icub-main and idyntree, both depending on yarp, will have yarp statically compiled. Not the best solution, but it should be fine.

In the **sdist case**, instead, either 1) the system has all the dependencies already installed, or 2) the pip packages contain all the headers, libs, and cmake imported targets.

This is to say that I don't have a real answer too. We can port this discussion somewhere else in the future and in the meantime merge this PR.

diegoferigo:
No wait, I think that the README should be changed since I substituted `dist-packages` with `site-packages`. And, thinking about what I wrote above, this PR is not portable since I have a virtualenv enabled. If the user system does not have a virtualenv enabled, yarp and idyntree install in `dist-packages`. This is a problem :/

traversaro:
> If the user system does not have a virtualenv enabled, yarp and idyntree install in dist-packages. This is a problem :/

That is the point, can't we just run the same script used by YARP and iDynTree here, and use that in the setup script?

diegoferigo:
We might, but having both entries in the scripts [as I pushed](https://github.com/robotology/robotology-superbuild/compare/2ed91c620d29667002ad3f770d878d3441538a1c..9d7ffbf7443c0d21392500e42a5b9eb69e9f2ae5) maybe it could be future proof on configurations we still didn't test, or if downstream projects do not use that command.

drdanz:
Sorry, but why don't we take this opportunity to drop support for python 2?
CMake 3.12 have a FindPython3 module

traversaro:
> 
> 
> We might, but having both entries in the scripts [as I pushed](https://github.com/robotology/robotology-superbuild/compare/2ed91c620d29667002ad3f770d878d3441538a1c..9d7ffbf7443c0d21392500e42a5b9eb69e9f2ae5) maybe it could be future proof on configurations we still didn't test, or if downstream projects do not use that command.

Ok, agreed.

traversaro:
> 
> 
> Sorry, but why don't we take this opportunity to drop support for python 2?
> CMake 3.12 have a FindPython3 module

Yes, but the problem is that  the minor version is still part of the installation prefix, so we should account for that.

Issue #366, opened 2020-03-17
Windows build is currently broken on Stable and Unstable branches


traversaro:
This is due to https://github.com/robotology/walking-controllers/issues/58 .

traversaro:
This probably was fixed by https://github.com/robotology/walking-controllers/pull/59 , I restarted the build.

traversaro:
The build still fails, see  https://github.com/robotology/walking-controllers/issues/58#issuecomment-600107149 .

traversaro:
Hi @GiulioRomualdi, if you want I can provide a PR to walking-controllers to fix this, let me know if it is ok.

GiulioRomualdi:
Working right now on it

traversaro:
> 
> 
> Working right now on it

Great, thanks!

traversaro:
Fixed by https://github.com/robotology/walking-controllers/pull/60 .

GiulioRomualdi:
If you want to update the `TAG` in the superbuild this is the release: https://github.com/robotology/walking-controllers/releases/tag/v0.3.2

traversaro:
> If you want to update the `TAG` in the superbuild this is the release: https://github.com/robotology/walking-controllers/releases/tag/v0.3.2

Thanks, the 2020.02 release is now done and we don't plan to update. The 0.3.2 can be a good candidate for the 2020.05 release.

PR #367, opened 2020-03-18
Update startup_robotology_superbuild.m.in

gabrielenava:
I've updated this file as follows:

- changed internal documentation (there were references to codyco and to mex-wholebodymodel);
- removed paths to `mex-wholebodymodel` (that are not used);
- added path to the [WBC](https://github.com/robotology/whole-body-controllers/tree/master/library/matlab-wbc/%2Bwbc) Matlab functions library, required for whole-body-controllers;
- added automatic generation of `goToWholeBodyControllers.m` script.

### DIFF
--- a/cmake/template/startup_robotology_superbuild.m.in
+++ b/cmake/template/startup_robotology_superbuild.m.in
@@ -1,22 +1,17 @@
 %% startup_robotology_superbuild.m
-%  Run this script only once to permanently add the required folders for using MATLAB toolbox (e.g. WBToolbox, whole-body-controllers) to your 
-%  MATLAB path. 
+%
+%  Run this script only once to permanently add the required folders for using
+%  MATLAB toolbox (e.g. WBToolbox, whole-body-controllers) to your MATLAB path. 
 
 fprintf('\nMATLAB Toolbox\n');
 
 installDir   = '@YCM_EP_INSTALL_DIR@';
-wbcSourceDir = ['@CMAKE_SOURCE_DIR@', '/robotology/whole-body-controllers/'];
 mexDir       = [installDir, filesep, 'mex'];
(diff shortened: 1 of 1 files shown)


traversaro:
Ideally, we should avoid references to the source directory, so everything continue to work only if we just keep the install  dir (as  it happens for example with the windows installer, even if it does not support Matlab at the moment).

traversaro:
Ignore the Windows failures, they are related to https://github.com/robotology/robotology-superbuild/issues/366 .

gabrielenava:
Follwing https://github.com/robotology/whole-body-controllers/issues/75#issuecomment-601622989, I've updated the startup file removing paths pointing to the source folder. The `wbc` library is now installed in the `mex` folder.

diegoferigo:
@gabrielenava Do you think useful also add the paths for the `wbc` matlab package also to the setup scripts? For example, in Ubuntu:

https://github.com/robotology/robotology-superbuild/blob/0cf19500a1ea733c4fa0bdcf56fcb3555ebcbac3/cmake/template/setup.sh.in#L47-L50

In this way people that just import the script in their bashrc will have WBC automatically configured.

traversaro:
> 
> 
> @gabrielenava Do you think it would be useful adding the path of the `wbc` matlab package also to the setup scripts? For example, in Ubuntu:
> 
> https://github.com/robotology/robotology-superbuild/blob/0cf19500a1ea733c4fa0bdcf56fcb3555ebcbac3/cmake/template/setup.sh.in#L47-L50
> 
> In this way people that just import the script in their bashrc will have WBC automatically configured.

As `+wbc` is already installed in `<prefix>/mex`, there is no need to add new paths to the `MATLABPATH`.

diegoferigo:
Ow ok, I missed that. Cool, ignore my comment than, thanks!

PR #368, opened 2020-03-19
README: Clarify initial windows dependency installation

traversaro:
What do you think about this @simeonedussoni ? If you think it is an improvement, feel free to approve the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,9 +245,7 @@ It is recommended to install the latest version of CMake.
 While this tool is not strictly required, it is convenient to install the [Rapid Environment Editor](https://www.rapidee.com/en/download) to easily modify the value of the [environment variables](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682653(v=vs.85).aspx) in Windows. 
 
 #### System Libraries
-The software in the superbuild depends on several C++  libraries that we assume are already available in your machine.
-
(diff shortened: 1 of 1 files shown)


simeonedussoni:
I feel more comfortable with no paragraph interruption between the two statements such as:
```
The software in the superbuild depends on several C++  libraries that we assume are already available in your machine otherwise you can install the required dependencies on your machineby using  [`vcpkg`] ...
```

in this way reader is aware that the arguments are "the same" 

Can you modify this then I will approve the PR (MODE "precisino" OFF)

traversaro:
> 
> 
> I feel more comfortable with no paragraph interruption between the two statements such as:
> 
> ```
> The software in the superbuild depends on several C++  libraries: you can install the required dependencies on your machine by using  [`vcpkg`] ...
> ```
> 
> in this way reader is aware that the arguments are "the same"
> 
> Can you modify this then I will approve the PR (MODE "precisino" OFF)

Good suggestion, thanks! Note that you can directly provide feedback directly by changing the code itself using PR suggestions, see https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request .

traversaro:
@simeonedussoni I hope I have addressed your comments.

PR #369, opened 2020-03-19
Add whole-body-estimators repository

traversaro:
Furthermore, do not install in codyco-modules anymore the devices that are now migrated in whole-body-estimators.

Unfortunately the Windows builds are still failing due to https://github.com/robotology/robotology-superbuild/issues/366, but we can  still check if Linux and macOs works fine. Furthermore, whole-body-estimators repo has its own Windows-based CI, so I think we should not risk regression w.r.t. to that. 

Note that the devices installed by codyco-modules will remain in `build/install` until the directory is removed. Given that `whole-body-estimators` installs its devices in `<prefix>/lib/yarp` the new device installed by `whole-body-estimators` should have the precedence. In any case (to avoid unplanned problems) **I suggest all users to just remove once the `build/install` directory of the superbuild, and then just populate it again by compiling the superbuild once**.


cmake/Buildwhole-body-estimators.cmake
```diff
@@ -0,0 +1,24 @@
...
+
+find_or_build_package(YARP QUIET)
+find_or_build_package(ICUB QUIET)
+find_or_build_package(iDynTree QUIET)
+
+set(whole-body-estimators_DEPENDS)
+
+list(APPEND whole-body-estimators_DEPENDS YARP)
+list(APPEND whole-body-estimators_DEPENDS ICUB)
+list(APPEND whole-body-estimators_DEPENDS iDynTree)
```
prashanthr05: Is there a way to handle the CMake option `IDYNTREE_USES_ICUB_MAIN` here? This is required for the compilation of `whole-body-estimators`. See https://github.com/robotology/whole-body-estimators/issues/46.

traversaro: Unfortunately there is no possibility of doing that. The iDynTree build by the superbuild is built with the `IDYNTREE_USES_ICUB_MAIN` , but if someone uses an external iDynTree with the IDYNTREE_USES_ICUB_MAIN set to OFF, there is no way to early fail at the moment.

prashanthr05: I see, thanks! . :+1:

cmake/Buildcodyco-modules.cmake
```diff
@@ -26,4 +26,8 @@ ycm_ep_helper(codyco-modules TYPE GIT
               COMPONENT dynamics
               FOLDER robotology
               DEPENDS ${codyco-module_DEPENDS}
-              CMAKE_ARGS -DCODYCO_USES_KDL:BOOL=OFF)
+              CMAKE_ARGS -DCODYCO_USES_KDL:BOOL=OFF
+                         -DENABLE_codycomod_genericSensorClient:BOOL=OFF
```
fjandrad: I see that here you disable the codyco devices. What remains from codyco that we still use ? Or is it an intermediate step before removing codyco?

traversaro: > 
> 
> I see that here you disable the codyco devices. What remains from codyco that we still use ? Or is it an intermediate step before removing codyco?

The open points are the yarpmotorgui files with the positions of the robot, see https://github.com/robotology/codyco-modules/issues/273 .

fjandrad: I see. I followed that discussion and thought it was solved I guess still has to be released. I just pinged on that.

PR #371, opened 2020-03-27
Boostrap and require recent YCM v0.11.1 in the superbuild

traversaro:
YCM v0.11.1 ( https://github.com/robotology/ycm/releases/tag/v0.11.1 ) contains some critical bug-fixes, so it is good to require and to use them.


HosameldinMohamed:
it looks like

> At least 1 approving review is required by reviewers with write acces

traversaro:
> 
> 
> it looks like
> 
> > At least 1 approving review is required by reviewers with write acces

Don't worry, any  review is ok, thanks!

PR #374, opened 2020-04-05
Remove codyco-modules, wholeBodyInterface and yarpWholeBodyInterface from robotology-superbuild

traversaro:
With the last bits of useful software  being migrated to whole-body-estimators and to whole-body-controllers (thanks @gabrielenava @prashanthr05  @HosameldinMohamed @fjandrad !) it is now the time to finally remove `codyco-modules` and its dependencies `wholeBodyInterface` and `yarpWholeBodyInterface`. 

To be completely honest, that feels a bit strange to me, as `codyco-modules` used to be the monolithic repository `codyco`, that it is also the first robotology repo that I maintained on GitHub. For example, it still contains the first iDynTree-related commit: 
* https://github.com/robotology/codyco-modules/commit/fe2a2068ebf43dd46020a91a64b966dd082b3638

Most of the software contained in the monolithic `codyco` repository  was moved to smaller repositories that were built together via the `codyco-superbuild` repo in 2014,  and so the `codyco` repository was renamed in `codyco-modules` . In 2017 then the codyco-superbuild was deprecated and replaced by the robotology-superbuild. 

We all learned a lot by building this software, especially by taking wrong decisions and understand practically why they were wrong. If anyone is interested, I strongly suggested to read  the paper "A Whole-Body Software Abstraction layer for Control Design of free-floating Mechanical Systems" available at https://web.archive.org/web/20190727180353/http://joser.unibg.it/index.php/joser/article/view/122/60 (apparently it is only available on Interent Archive as the [JOSER (the journal where the paper was submitted) website](http://www.joser.org/) is now down -_- ), in particular the Lessons Learned section.  

To celebrate a bit the moment I will also ask a celebratory review from @francesco-romano . : )


traversaro:
fyi @iron76 @DanielePucci @jeljaik : )

DanielePucci:
This is definitely much worse than https://github.com/robotology/codyco-modules/issues/290

goodbye

prashanthr05:
Just a curiosity. Do you think it would make sense to have an `unreleased.yaml` in releases folder to already incorporate these changes? 

Maybe also, we could add a line in the documentation about,
- these changes and need to compile codyco-modules  independently (if someone is still using it) and not from the superbuild. 
- possible clashes in the installed plugins of whole-body-estimators and previously existing codyco-modules

traversaro:
> these changes and need to compile codyco-modules independently (if someone is still using it) and not from the superbuild.

`codyco-modules` is going to the software paradise, no need for further documentation. . : )

> possible clashes in the installed plugins of whole-body-estimators and previously existing codyco-modules

I think it would be just confusing, as it does not apply at all to new installations. If anyone has problems of that kind, he can open a new issue to receive promp support, but I would not add mmore details on the docs that is already quite confusing for new users.

> Just a curiosity. Do you think it would make sense to have an unreleased.yaml in releases folder to already incorporate these changes?

Yes, I am going to do something similar to that soon.

PR #386, opened 2020-04-25
Add preliminary support for 2020.05 release

traversaro:
* Add file containing versions of projects for 2020.05 release  (for now just a copy of 2020.02's tags, with just the change of whole-body-estimators and  the removal of codyco-modules and deps to account for https://github.com/robotology/robotology-superbuild/pull/374)
* Add support for testing in CI a release (2020.05 for now) and also test the default `ROBOTOLOGY_PROJECT_TAGS` setting instead of hardcode `Stable` (so we actually test correctly the `releases/**` branches

### DIFF
--- a/releases/2020.05.yaml
+++ b/releases/2020.05.yaml
@@ -1,0 +1,125 @@
+repositories:
+  qpOASES:
+    type: git
+    url: https://github.com/robotology-dependencies/qpOASES.git
+    version: v3.2.0.1
+  osqp:
+    type: git
+    url: https://github.com/oxfordcontrol/osqp.git
+    version: v0.6.0
+  ycm:
+    type: git
+    url: https://github.com/robotology/ycm.git
+    version: v0.11.1
+  YARP:
+    type: git
+    url: https://github.com/robotology/yarp.git
+    version: v3.3.2
+  ICUB:
+    type: git
+    url: https://github.com/robotology/icub-main.git
+    version: v1.15.0
+  ICUBcontrib:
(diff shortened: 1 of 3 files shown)


releases/2020.05.yaml
```diff
@@ -0,0 +1,125 @@
...
+    url: https://github.com/robotology/robots-configuration.git
+    version: v1.15.0
+  GazeboYARPPlugins:
+    type: git
+    url: https://github.com/robotology/gazebo-yarp-plugins.git
+    version: v3.3.0
+  icub-gazebo:
+    type: git
+    url: https://github.com/robotology/gazebo-yarp-plugins.git
+    version: v1.15.0
```
diegoferigo, line 37: wrong url

traversaro: Thanks, fixed in https://github.com/robotology/robotology-superbuild/pull/386/commits/a7fc0dda4ae275468cc18f36250d25355d0d1f27 .

traversaro:
The workflow to modify the 2020.05.yml file is explained in https://github.com/robotology/robotology-superbuild/issues/387 .

traversaro:
The build is currently broken by the fact that WB-Toolbox 5 that uses deprecated YCM features, but that are actually broken. The problem was fixed  in  Stable branches in https://github.com/robotology/wb-toolbox/pull/184, but it was never released.

I opened the issue https://github.com/robotology/wb-toolbox/issues/187 to ask for a new release to @diegoferigo .

Issue #121, opened 2018-09-24
CMake error when trying to compile the superbuild with Yarp


gabrielenava:
I downloaded from scratch the `robotology-superbuild` on my laptop and I tried to compile it. I do not have `Yarp` installed on my lapotop, it is supposed to be installed together with the superbuild. Nevertheless, when I tried to compile the project I got the following error:

```CMake


 CMake Error at /home/gnava/Software/github/INSTALL/lib/ICUBcontrib/ICUBcontribConfig.cmake:7 (find_package):
   By not providing "FindYARP.cmake" in CMAKE_MODULE_PATH this project has
   asked CMake to find a package configuration file provided by "YARP", but
   CMake did not find one.

   Could not find a package configuration file provided by "YARP" with any of
   the following names:

     YARPConfig.cmake
     yarp-config.cmake

   Add the installation prefix of "YARP" to CMAKE_PREFIX_PATH or set
   "YARP_DIR" to a directory containing one of the above files.  If "YARP"
   provides a separate development package or SDK, be sure it has been
   installed.
 Call Stack (most recent call first):
   build/install/share/YCM/modules/FindOrBuildPackage.cmake:180 (find_package)
   CMakeLists.txt:111 (find_or_build_package)

```

Maybe some path got broken recently?

##### I have Ubuntu 16.04 as OS

traversaro:
I guess you have an external installation of ICUBcontrib install as `CMAKE_INSTALL_PREFIX` :  `/home/gnava/Software/github/INSTALL/`, that was configured with a YARP installation that is not anymore installed or not anymore present in the `CMAKE_PREFIX_PATH`.

gabrielenava:
>I guess you have an external installation of ICUBcontrib install as CMAKE_INSTALL_PREFIX : /home/gnava/Software/github/INSTALL/, that was configured with a YARP installation that is not anymore installed or not anymore present in the CMAKE_PREFIX_PATH.

Yes, I have it. Ok then I will address this on my pc thanks!

traversaro:
Exactly, ICUBcontrib is now a project handled in the superbuild, so you either: 
* Install it with the superbuild 
or 
* use it externally, but all its dependencies should also be installed externally (in this case, YARP)

Issue #92, opened 2018-06-19
Superbuild finds its own subproject components on Ubuntu if <package>_DIR env variables are defined


traversaro:
~~~
icub@icub-head:/usr/local/src/robot/robotology-superbuild/build$ make
-- YCM not found. Bootstrapping it.
-- Downloading YCMEPHelper.cmake - Attempt 1 of 3
-- Downloading YCMEPHelper.cmake - SUCCESS
-- Downloading StyleGITHUB.cmake - Attempt 1 of 3
-- Downloading StyleGITHUB.cmake - SUCCESS
-- Performing download step (git clone) for 'YCM'
-- Performing update step for 'YCM'
-- Performing configure step for 'YCM'
-- Performing uninstall step for 'YCM'
-- Performing build step for 'YCM'
-- Performing install step for 'YCM'
-- Found YCM: /usr/local/src/robot/robotology-superbuild/build/install/share/YCM (found version "0.5.1.20171220.1+gite9bfa39")
-- Package RTF not found. Will be downloaded and built.
-- Using iCub from install
-- Package iDynTree not found. Will be downloaded and built.
-- Package wholeBodyInterface not found. Will be downloaded and built.
-- Package yarpWholeBodyInterface not found. Will be downloaded and built.
-- Package codyco-modules not found. Will be downloaded and built.
CMake Deprecation Warning at build/install/lib/cmake/YARP/YARPConfig.cmake:272 (message):
  The YARP_MODULE_PATH variable is deprecated.  CMake find package modules
  are now in YCM.
Call Stack (most recent call first):
  build/install/lib/ICUBcontrib/ICUBcontribConfig.cmake:9999 (_yarp_module_path_is_deprecated)
  build/install/share/YCM/modules/FindOrBuildPackage.cmake:180 (find_package)
  cmake/Buildrobots-configuration.cmake:9 (find_or_build_package)
  build/install/share/YCM/modules/FindOrBuildPackage.cmake:235 (include)
  CMakeLists.txt:152 (find_or_build_package)


CMake Error at build/install/lib/ICUBcontrib/ICUBcontribConfig.cmake:12 (include):
  include could not find load file:

    YarpInstallationHelpers
Call Stack (most recent call first):
  build/install/share/YCM/modules/FindOrBuildPackage.cmake:180 (find_package)
  cmake/Buildrobots-configuration.cmake:9 (find_or_build_package)
[...]

PR #389, opened 2020-04-26
Set YCM option YCM_DISABLE_SYSTEM_PACKAGES to ON

traversaro:
During the past years, we experienced quite a lot of times that for some reason on a setup 
the superbuild was finding its own installed packages, or finding the one of a similar superbuild 
installed on another setup. As recovering from this kind of errors is tricky for users, and as 
most of robotology-superbuild users do not need to find  the packages installed by the superbuild in  their system, the YCM_DISABLE_SYSTEM_PACKAGES option was introduced in YCM 
( see https://github.com/robotology/ycm/pull/332 )  to disable the use of  any system installed package if the same package can be installed by the superbuild (equivalent of setting  all the USE_SYSTEM_<project> options to OFF). 

This option is set by default to `OFF` on YCM for backward compatibility, but to avoid all the problems in the superbuild we set it to `ON` by default. If an expert user want to be able to find system-installed packages (for example because he has a  system-installed YARP) he just need to set it at `OFF`. 

The option is still not supported in the stable released version of YCM, so I will not update the README for now, but  it will be already effective if a users specifies uses `Unstable` `ROBOTOLOGY_PROJECT_TAGS`, so for example it  will be  already useful  for the iCubGenova01 setup @randaz81 @xEnVrE .

Related issues: 
* https://github.com/robotology/robotology-superbuild/issues/121
* https://github.com/robotology/robotology-superbuild/issues/174
* https://github.com/robotology/robotology-superbuild/issues/92

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,8 @@ include(FeatureSummary)
 
 set(YCM_USE_CMAKE_PROPOSED TRUE CACHE BOOL "Use files including unmerged cmake patches")
 
-# Compilation options
+# YCM options 
+option(YCM_DISABLE_SYSTEM_PACKAGES "Disable use of all the system installed packages" ON)
 
 # Dependencies options 
 ## Matlab related options


lrapetti:
> If an expert user want to be able to find system-installed packages (for example because he has a system-installed YARP) he just need to set it at `ON`.

Here you meant `OFF` right?

traversaro:
> 
> 
> > If an expert user want to be able to find system-installed packages (for example because he has a system-installed YARP) he just need to set it at `ON`.
> 
> Here you meant `OFF` right?

Yes, sorry for the typo, it should be fixed now.

traversaro:
@lrapetti if you think that this PR make sense, feel free to approve it, thanks!

Issue #29, opened 2018-01-30
Add "check your installation" README section for ROBOTOLOGY_USES_OCTAVE option 


traversaro:
Once it is documented, we can remove the `mark_as_advanced`.

PR #392, opened 2020-04-26
Update Matlab and  Octave instructions with "Check the installation" section

traversaro:
This is useful for users to perform a basic sanity check of their installation on their own.

Fix https://github.com/robotology/robotology-superbuild/issues/32 .
Fix https://github.com/robotology/robotology-superbuild/issues/29 .


traversaro:
Friendly ping @gabrielenava

Issue #356, opened 2020-02-25
Disable install target


traversaro:
As discussed in https://github.com/robotology/ycm/issues/252, this cannot be handled at the YCM level, so we should handle in this repo, as simply ignoring the install target in the docs turned out to be not sufficient to avoid problems.

PR #395, opened 2020-04-27
 Disable YCM-created install target

traversaro:
The `install` target of the superbuild is not supported for several reason, and users that want to install their project in a specific directory can use  the existing YCM variable YCM_EP_INSTALL_DIR  (see https://github.com/robotology/ycm/pull/289 ).

However, to avoid that some user that does not read the docs runs the install target, let's disable it. 

Fix https://github.com/robotology/robotology-superbuild/issues/356 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,10 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 include(CTest)
 include(FeatureSummary)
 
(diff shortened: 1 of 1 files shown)


traversaro:
@bemilio I think you are the most qualified person to review this PR. : )

bemilio:
If only this was there some months ago :( - or if only the younger me had more patience in reading the docs

Also according to [this comment])https://github.com/robotology/ycm/issues/252#issuecomment-491713290) it looks a good solution, approved

PR #399, opened 2020-05-01
Migrate YCMBootstrap to use CMake's FetchContent 

traversaro:
The YCMBootstrap process has been historically quite error-prone. This PR substitute this process by a more straightforward use of the CMake's FetchContent  repo. 

With respect to the past, the main difference is the superbuild will not use exactly the same `YCM` downloaded in `robotology-superbuild/robotology/YCM` , as a private copy of YCM used only for the robotology-superbuild configuration will be downloaded in `build/_deps/ycm-src`  . However, as the new script still observers the `YCM_TAG`  variable, the version downloaded in `build/_deps/ycm-src`  will be exactly the same version of YCM that will be downloaded in  `robotology-superbuild/robotology/YCM` .

On my Windows machine, this decreases the time spent in the YCM  bootstrap to ~5 seconds from ~20 seconds of the previous procedure. 

If this new module will work fine, it will be proposed in YCM to fix  https://github.com/robotology/ycm/issues/138 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,7 +119,7 @@ endif()
 set(YCM_FOLDER robotology)
 set(YCM_COMPONENT core)
 set(YCM_MINIMUM_VERSION 0.11.1)
-include(YCMBootstrap)
+include(YCMBootstrapFetchContent)
 
 include(FindOrBuildPackage)
 include(YCMEPHelper)
--- a/cmake/BuildICUB.cmake
+++ b/cmake/BuildICUB.cmake
@@ -5,9 +5,11 @@
 include(YCMEPHelper)
 include(FindOrBuildPackage)
 
+find_or_build_package(YCM QUIET)
 find_or_build_package(YARP QUIET)
 
 set(ICUB_DEPENDS "")
+list(APPEND ICUB_DEPENDS YCM)
 list(APPEND ICUB_DEPENDS YARP)
 
 if(ROBOTOLOGY_ENABLE_ICUB_HEAD)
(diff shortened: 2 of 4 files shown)


traversaro:
> Furthermore, this PR works fine unless YCM functionality is needed in the top-level CMake file. It does not seem to be the case though. Otherwise, you should have used `PARENT_SCOPE` when extending `CMAKE_MODULE_PATH`.

The boostrap file is already  included in the top-level CMake file, so I don't think it is  need to specify the `PARENT_SCOPE` when extending `CMAKE_MODULE_PATH` .

traversaro:
For some reason that I don't get the CI is not starting on this PR.

traversaro:
> On my Windows machine, this decreases the time spent in the YCM bootstrap to ~5 seconds from ~20 seconds of the previous procedure.

This seems to be confirmed in the CI tests, in which the configuration time on Windows passed from ~1:20 seconds to ~55 seconds.

traversaro:
It turns out that consuming `YCMEPHelper` directly is  quite difficult, as  `YCMEpHelper` has a check for `YCM_FOUND` in its code, and if  `YCM_FOUND` is not found, it uses bootstrap-like behaviour (downloading stuff over the internet): https://github.com/robotology/ycm/blob/master/modules/YCMEPHelper.cmake#L106 . Setting `YCM_FOUND` for fooling that check is not feasible, as it will prevent the superbuild to correctly setup the use of YCM as a subproject. So it is necessary first of all to modify  `YCMEPHelper` .

traversaro:
Closing this PR as it is not possible to achieve this without modifying YCM itself.

traversaro:
> Setting `YCM_FOUND` for fooling that check is not feasible, as it will prevent the superbuild to correctly setup the use of YCM as a subproject. So it is necessary first of all to modify `YCMEPHelper` .

I actually did that in https://github.com/robotology/robotology-superbuild/pull/1078 and it is working fine.

Issue #335, opened 2020-01-29
Add a GitHub Action job to build a Windows installer on a new tag


traversaro:
Using the scripts provided in https://github.com/robotology/robotology-superbuild/pull/334 .

Useful resources: 
* GitHub Actions documentation: :
  - https://help.github.com/en/actions/automating-your-workflow-with-github-actions
* Example GitHub Actions workflow for a simple CMake/C++ project :  
  - https://github.com/robotology/how-to-export-cpp-library/blob/master/.github/workflows/ci.yml
* GitHub Action to upload a binary to a release: 
  -  https://github.com/actions/upload-release-asset
* Existing script for building on Windows (it is quite complex to support multiplatform, the one for building windows installers can be simpler)
  - https://github.com/robotology/robotology-superbuild/blob/master/.github/workflows/ci.yml

traversaro:
Related PR: https://github.com/robotology/robotology-superbuild/pull/351 .

traversaro:
Related PR: https://github.com/robotology/robotology-superbuild/pull/351 .

traversaro:
Example of use of the upload-release-asset action: https://github.com/bradleysepos/mingw-w64-build/pull/4/files#diff-898e737ee4e56ba042bda31764bed49eR85 .

PR #404, opened 2020-05-03
Add generation and upload of Windows installer in GitHub Actions script

traversaro:
To avoid duplication, we reuse most of the existing logic of the Windows CI, the only modifications  are: 
* We always set (in Windows) the `-DYCM_EP_INSTALL_DIR=C:/robotology/robotology`, to match what we will use in the installer generation 
* For builds associated to a release event, we also build the superbuild in Debug mode, so that the Debug libraries are included in the installer 
* We generate the installer using the instructions added in https://github.com/robotology/robotology-superbuild/pull/334
* We then use the [`actions/upload-release-asset`](https://github.com/actions/upload-release-asset) Action to upload the generated installer to the Release page (I added a `ACCESS_TOKEN` Secret to the repo containing an appropriate token) 

Fix https://github.com/robotology/robotology-superbuild/issues/335 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,8 @@ on:
     tags:
       - v*
   pull_request:
+  release:
+    types: [published]
   schedule:
   # * is a special character in YAML so you have to quote this string
   # Execute a "nightly" build at 2 AM UTC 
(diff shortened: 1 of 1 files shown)


traversaro:
An example of installer generated by this script and uploaded to a release is available at https://github.com/traversaro/robotology-superbuild/releases/tag/test-installer-v4 .

traversaro:
Some questions from @Nicogene : 

> 1) Where is defined the target `PACKAGE`?

The `PACKAGE` CMake target is created by the `CPack` module, see https://cmake.org/cmake/help/latest/module/CPack.html . The target is all capital because of Visual Studio (I guess) on other generators it will probably be  `package` .

> 2) I saw that you used https://doc.qt.io/qtinstallerframework/index.html for creating the installer, but you designed also the GUI of the installer?

To clarify, I used `qtinstallerframework` via the corresponding CPack generator (see https://gitlab.kitware.com/cmake/community/-/wikis/doc/cpack/Packaging-With-CPack and https://cmake.org/cmake/help/v3.17/module/CPackIFW.html), so I did not use  `qtinstallerframework` directly. More in general, I think that QtIFW can be customized via scripts (as the one that we  use to modify the environment variables) but I think that the basic GUI is fixed.

PR #408, opened 2020-05-15
 Add missing dependencies of iDynTree on OsqpEigen 

traversaro:
Before this fix, whetever iDynTree was compiled or not with OsqpEigen support was non 
deterministic, and based on wheter OsqpEigen  was already compiled and  installed when
iDynTree was configured. With  this fix, the dependency of iDynTree on OsqpEigen is explicitly 
indicated, and the IDYNTREE_USES_* options are explicity set, so that if for some reason one
of the dependency is not installed correctly, iDynTree configuration will fail.

Also add Add fail-fast: false also to Docker-based GitHub Actions jobs, that had the nice side effect: of helping in debug the problems discussed in https://github.com/robotology/osqp-eigen/pull/62 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 This is a meta repository (so-called "superbuild") that uses [CMake](https://cmake.org/) and [YCM](https://github.com/robotology/ycm) to automatically
 download and compile software developed in the robotology GitHub organization, such as the YARP middleware or software used to run the iCub humanoid robot.
 
+
 [CMake](https://cmake.org/) is an open-source, cross-platform family of tools designed to build, test and package software.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -3,6 +3,7 @@
 This is a meta repository (so-called "superbuild") that uses [CMake](https://cmake.org/) and [YCM](https://github.com/robotology/ycm) to automatically
 download and compile software developed in the robotology GitHub organization, such as the YARP middleware or software used to run the iCub humanoid robot.
 
+
```
S-Dafarra, line 6:
```suggestion
```

PR #412, opened 2020-05-19
Cleanup Windows packaging code

traversaro:
* Reuse as much as possible the scripts already provided by https://github.com/robotology/robotology-superbuild-dependencies-vcpkg 
* Add support to install Gazebo as part of the dependencies installer and setup all the required environment variables 
* In CI also generate the dependencies installer 

I also cleaned up the handling of components, so now it is possible to use the full installer to just install dependencies, by deselecting the installation of robotology software. Furthermore, I also fixed the problem that prevented the deselection of the "Update Enviroment Variables" to manage them manually, that was reported by @randaz81 some time ago.


traversaro:
If someone wants to give a spin to the generated installers, he/she can find it here: https://github.com/traversaro/robotology-superbuild/releases/tag/test-pr412-v1 .

PR #414, opened 2020-05-21
GitHub Actions: Use gazebo11 on macOS

traversaro:
Using the latest version make sense in general (it is the version that new users will install by default) and will also help as a workaround against https://github.com/osrf/homebrew-simulation/issues/1022 .


traversaro:
Apparently also ignition-msgs5 is still affected by https://github.com/osrf/homebrew-simulation/issues/1022 .

traversaro:
It actually seems that now `gazebo10` (as used in `master` branch) is working fine, while `gazebo11` has still problems to be installed.

traversaro:
On Gazebo11 everything seems to work, so I will merge this PR.  cc @nunoguedelha

PR #416, opened 2020-05-26
Add blocktest to the robotology-superbuild

Nicogene:
This is a first draft for adding `blocktest` and `blocktest-yarp-plugins` to the `robotology-superbuild`.

For sure it misses:

- [x] CI
- [x] Documentation


traversaro:
I would strongly prefer to avoid having single dependency components, as it increase complexity for CI, documentation and  in general user support. As `blocktest`  is related to robot testing, do you think it could make sense to add it to the existing `ROBOTOLOGY_ENABLE_ROBOT_TESTING` component? To clarify, the `ROBOT_TESTING` in the component name is a generic term, it is not relative specifically to the `robot-testing-framework` .

Nicogene:
For me is fine, I was worried only for the ambiguity of the name and the differences of dependecies of the 2 repository. My main concern is about boost, it would require to install boost if you want to compile `robot-testing-framework` via superbuild am I right?
I totally agree about the reduction of complexity

traversaro:
>  My main concern is about boost, it would require to install boost if you want to compile robot-testing-framework via superbuild am I right? I totally agree about the reduction of complexity

I think the problem created by this downside is minor w.r.t. to the advantages by having just one component.

traversaro:
In the meanwhile, I think if one of these two repos have an unstable (`devel` ) branch, you can add it in `cmake/ProjectTagsUnstable.cmake` .

Nicogene:
I added it under the ROBOT_TESTING profile, in this way both documentation and CI should be ok, am I right?

traversaro:
> 
> 
> I added it under the ROBOT_TESTING profile, in this way both documentation and CI should be ok, am I right?

Ok for the CI, for the docs probably you may want to update the table in https://github.com/robotology/robotology-superbuild#profile-cmake-options  ?

README.md
```diff
@@ -386,7 +386,7 @@ The steps necessary to install the system dependencies of the Robot Testing prof
 operating system-specific installation documentation, and no additional required system dependency is required.
 
 ### Check the installation
-If the profile has been correctly enabled and compiler, you should be able to run the `robottestingframework-testrunner` executable from the command line.
+If the profile has been correctly enabled and compiler, you should be able to run the `robottestingframework-testrunner` and `blocktestrunner`executable from the command line.
```
traversaro, line 389:
```suggestion
If the profile has been correctly enabled and compiler, you should be able to run the `robottestingframework-testrunner` and `blocktestrunner` executable from the command line.
```

traversaro:
CI passed at the first trial, nice!

traversaro:
Thanks  @Nicogene !

PR #423, opened 2020-06-30
Create 2020.08.yaml for 2020.08 release

traversaro:
It contains the tag as released in 2020.05, with some projects bumped to use thee branches on which the release contained in 2020.08 will be tagged.

See https://github.com/robotology/robotology-superbuild/blob/master/doc/developers-faqs.md#how-to-do-a-new-release for more details.

### DIFF
--- a/releases/2020.08.yaml
+++ b/releases/2020.08.yaml
@@ -50,11 +50,11 @@ repositories:
   blockTest:
     type: git
     url: https://github.com/robotology/blocktest.git
-    version: v2.2.0
+    version: d490bd018cf6c40e0a2e37650943d43de23751a0
   blocktest-yarp-plugins:
     type: git
     url: https://github.com/robotology/blocktest-yarp-plugins.git
-    version: v1.0.0
+    version: 78eacbf220a1c6db6d1bac68e4b51ab4a3a55142
   iDynTree:
     type: git
     url: https://github.com/robotology/idyntree.git
--- a/releases/2020.08.yaml
+++ b/releases/2020.08.yaml
@@ -58,7 +58,7 @@ repositories:
   iDynTree:
     type: git
     url: https://github.com/robotology/idyntree.git
-    version: v1.0.2
+    version: v1.1.0
   BlockFactory:
     type: git
     url: https://github.com/robotology/blockfactory.git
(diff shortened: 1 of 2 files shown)


traversaro:
@Nicogene The latest release of `blocktest-yarp-plugins` is failing with these errors: 
~~~
 CMake Error at CMakeLists.txt:57 (add_library):
  Target "blocktestyarpplugins" links to target "Boost::system" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?


CMake Error at CMakeLists.txt:57 (add_library):
  Target "blocktestyarpplugins" links to target "Boost::filesystem" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
~~~
What is the plan for 2020.08 for `blocktest-yarp-plugins`  ?

Nicogene:
`blocktest-yarp-plugins` latest release is quite outdated(July 2019), for sure we need a new release, and maybe also for `blocktest`.

It would be great if we could make `blocktest*` work with the superbuild and then add this feature in the release.

cc @triccyx

traversaro:
> 
> 
> `blocktest-yarp-plugins` latest release is quite outdated(July 2019), for sure we need a new release, and maybe also for `blocktest`.
> 
> It would be great if we could make `blocktest*` work with the superbuild and then add this feature in the release.
> 
> cc @triccyx

For the time being, I used the latest commit in the `master` branches of both repos.

releases/2020.08.yaml
```diff
@@ -0,0 +1,133 @@
...
+    url: https://github.com/robotology/icub-tests.git
+    version: v1.16.0
+  blockTest:
+    type: git
+    url: https://github.com/robotology/blocktest.git
+    version: d490bd018cf6c40e0a2e37650943d43de23751a0
+  blocktest-yarp-plugins:
+    type: git
+    url: https://github.com/robotology/blocktest-yarp-plugins.git
+    version: 78eacbf220a1c6db6d1bac68e4b51ab4a3a55142
```
Nicogene, line 57: It wasn't sufficient writing `master`?

traversaro: Ideally the release file should never contain branches, but always fixed tags or commits to ensure that we are always ready to do a release. As a temporary workaround we have branches for yarp/icub due to the fact that 3.4 and 1.16 should happen before august, but given that for blockTest I have no idea of when the release will be done, I prefer to played safe.

PR #424, opened 2020-07-06
setup.sh/bat: add BLOCKTEST_RESOURCE_PATH

Nicogene:
This env var is needed to make work `blocktestrunner` and `ScriptBuilder` from any directory.

WIP until https://github.com/robotology/blocktest/pull/35 is not merged


Nicogene:
@traversaro I successfully tested the changes of this PR and the ones in `blocktest` have been merged, if it is ok for you I would merge

traversaro:
To  make the installer  works correctly, you need also to update the powershell scripts https://github.com/robotology/robotology-superbuild/blob/master/cmake/template/addPathsToUserEnvVariables.ps1.in and https://github.com/robotology/robotology-superbuild/blob/master/cmake/template/removePathsFromUserEnvVariables.ps1.in .

Nicogene:
One build is failing but I don't understand which is the problem

traversaro:
> 
> 
> One build is failing but I don't understand which is the problem

That failure was already  in the base branch and it was fixed in https://github.com/robotology/robotology-superbuild/pull/423 , so you can ignore that. However, can you also update the docs in https://github.com/robotology/robotology-superbuild/blob/450c44dedacfa881122e52d59b0807dcb21deaea/doc/environment-variables-configuration.md#robot-testing ? I promise it is the last step (probably we should add  a note on dev faqs on how to handle a new env variable).

PR #436, opened 2020-07-27
Add diagnostic daemon

Nicogene:
This PR add the `diagnostidaemon` to the `ICUB_HEAD` profile.

It will be draft(and I think not passing CI) until `diagnostic-daemon` won't be public


CMakeLists.txt
```diff
@@ -195,6 +195,9 @@ endif()
 if(ROBOTOLOGY_ENABLE_ICUB_HEAD)
   find_or_build_package(icub-firmware-shared)
   find_or_build_package(ICUB)
+  if(NOT APPLE)
```
traversaro, line 198: Can we at least give it a shot to compile it under `macOS` ? Otherwise, we need to explicitly document that `diagnosticdaemon` is not compile in macOS in the README, but if we can avoid to do this it would be great.

Nicogene:
@traversaro the build of `macOS` is failing due to : 
```
/diagnosticdaemon.build/Release/diagnosticdaemon.build/Objects-normal/x86_64/diagnosticdaemon_dependency_info.dat -o /Users/runner/work/robotology-superbuild/robotology-superbuild/build/robotology/diagnosticdaemon/Release/diagnosticdaemon
2020-07-28T12:51:20.1520890Z ld: library not found for -lstdc++fs
2020-07-28T12:51:20.1521900Z clang: error: linker command failed with exit code 1 (use -v to see invocation)
2020-07-28T12:51:20.1523300Z 
2020-07-28T12:51:20.1555710Z ** BUILD FAILED **
2020-07-28T12:51:20.1557240Z 
```

I added that library for using `<experimental/filesystem>` since `<filesystem>` is available only since `gcc 8` and `Ubuntu 18.04` has `gcc 7.5`.

It is possible to add the macOs compatibility but I don't think it would be worthy. 
Once we will drop `Ubuntu 18.04` we will use `<filesystem>` and we should have the compatibility.

Is it ok for you if for now I add `if(NOT APPLE)` in `CMakeLists.txt`?

traversaro:
> Is it ok for you if for now I add `if(NOT APPLE)` in `CMakeLists.txt`?

Ok, please add a comment mention that this is necessary because of https://github.com/robotology/robotology-superbuild/issues/439 .

traversaro:
Hi @Nicogene, this PR broke the CI, in particular on Release 2020.08, can you check it? In general when I "Approve" a PR I mean as "We can merge it as long as the CI passes".

traversaro:
> 
> 
> Hi @Nicogene, this PR broke the CI, in particular on Release 2020.08, can you check it? In general when I "Approve" a PR I mean as "We can merge it as long as the CI passes".

Actually I was wrong, the regression in there is not related to that PR, see https://github.com/robotology/robotology-superbuild/issues/441 .

PR #437, opened 2020-07-27
blocktest: fix with the right package name

Nicogene:
The right name of the package is `blocktestcore`.

please review code

### DIFF
--- a/cmake/Buildblocktestcore.cmake
+++ b/cmake/Buildblocktestcore.cmake
@@ -4,7 +4,7 @@
 
 include(YCMEPHelper)
 
-ycm_ep_helper(blockTest TYPE GIT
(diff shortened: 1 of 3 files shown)


cmake/Buildblocktestcore.cmake
```diff
@@ -4,9 +4,8 @@
 
 include(YCMEPHelper)
 
-ycm_ep_helper(blockTest TYPE GIT
+ycm_ep_helper(blocktestcore TYPE GIT
               STYLE GITHUB
               REPOSITORY robotology/blocktest.git
               TAG master
-              COMPONENT core
```
traversaro, line 11: Please leave this line, as it is not related to the name of the CMake package at all.

Issue #440, opened 2020-07-31
GitHub Actions macOS build is failing


traversaro:
For the usual homebrew update problems:
~~~
2020-07-31T02:19:59.4766700Z ==> Installing osrf/simulation/gazebo11
2020-07-31T02:20:15.2296560Z ==> Patching
2020-07-31T02:20:15.2297100Z ==> Applying brew_python_fix.patch
2020-07-31T02:20:15.2430950Z patching file deps/opende/CMakeLists.txt
2020-07-31T02:20:15.2475190Z ==> cmake .. -DQWT_WIN_INCLUDE_DIR=/usr/local/lib/qwt.framework/Headers -DQWT_WIN_LIBRARY_DIR=/usr/local/lib/qwt.framework
2020-07-31T02:21:11.2612540Z ==> make install
2020-07-31T02:45:52.7464010Z Last 15 lines from /Users/runner/Library/Logs/Homebrew/gazebo11/02.make:
[...]

PR #443, opened 2020-08-07
Update startup_robotology_superbuild.m

gabrielenava:
I added the path to the folder where I install the simulink models coming from the new [whole-body-controllers Simulink library](https://github.com/robotology/whole-body-controllers/tree/devel/library/simulink-library) that i set up. For the moment the library is available in WBC `devel` branch, but adding already the path shouldn't be a problem because `startup_robotology_superbuild.m` simply ignores not existing paths.


traversaro:
Do you think that `addpath(genpath(mex))` is an overkill, so downstream projects can just install there without the need to updating everytime a file in the superbuild?

gabrielenava:
>Do you think that addpath(genpath(mex)) is an overkill, so downstream projects can just install there without the need to updating everytime a file in the superbuild?

Normally is a bit strong, but on the other hand I don't see why someone in the future may install inside `mex` a folder that is not supposed to be added to the MATLAB path. To me we can also use `genpath`.

gabrielenava:
modified with `genpath` and tested on my laptop. Seems working fine.

traversaro:
On the other hand, probably if we do `genpath` we are also adding to the path the content of `+iDynTree` and similar "namespace" directories, and I guess this is not desirable?

gabrielenava:
I checked on my laptop and it seems that even if the package folders are directly added to the path, MATLAB is smart enough to not allow you to call the functions inside the package without using the proper namespace. 

I thought that maybe a less invasive solution could be to create a dedicated folder inside `mex` that we addpath(genpath()), and then install in this dedicated folder.

traversaro:
> I checked on my laptop and it seems that even if the package folders are directly added to the path, MATLAB is smart enough to not allow you to call the functions inside the package without using the proper namespace.

Cool, in that case I guess the proposed solution is good to go.

gabrielenava:
From https://www.mathworks.com/help/matlab/matlab_oop/scoping-classes-with-packages.html

>Packages and the MATLAB Path
>
>You cannot add package folders to the MATLAB path, but you must add the package parent folder to the MATLAB path. Package members are not accessible if the package parent folder is not on the MATLAB path, even if the package folder is the current folder. Making the package folder the current folder is not sufficient to add the package parent folder to the path.

so our solution should be ok!

traversaro:
> 
> 
> > I checked on my laptop and it seems that even if the package folders are directly added to the path, MATLAB is smart enough to not allow you to call the functions inside the package without using the proper namespace.
> 
> Cool, in that case I guess the proposed solution is good to go.

Sorry, I reviewed a bit our docs on MATLAB configuration (see https://github.com/robotology/robotology-superbuild#matlab) and `startup_robotology_superbuild.m` is just one of the many way in which we support to set the MATLAB path, so we can't just change `startup_robotology_superbuild.m` and leave the other ways inconsistent. As some other ways can't use `genpath`, I guess for nowwe just go back to add the path you neeed to add, and then we can look in thefuture how to make this more mantainable.

gabrielenava:
I reverted to the original solution, and we'll think how to handle it better in a future release.

traversaro:
> 
> 
> I reverted to the original solution, and we'll think how to handle it better in a future release.

Thanks @gabrielenava . To avoid inconsistencies for users, we should update all the points in which the MATLAB required paths are either documented or automatically added, beside the file you already modified at a first glance this means: 
* The docs in https://github.com/robotology/robotology-superbuild#start-matlab-from-the-launcher-or-the-application-menu
* The command prompt setup script in https://github.com/robotology/robotology-superbuild/blob/095bd1a124ded942d8c0e544a42db9badeee04d6/cmake/template/setup.bat.in#L42 
* The bash setup script in https://github.com/robotology/robotology-superbuild/blob/095bd1a124ded942d8c0e544a42db9badeee04d6/cmake/template/setup.sh.in#L47

traversaro:
Please ignore the macOS failures, they are due to https://github.com/robotology/robotology-superbuild/issues/440 .

gabrielenava:
>Thanks @gabrielenava ! Can I squash the commits in a single one?

Sure!

PR #446, opened 2020-08-18
Add modifications for 2020.05.patch-01 release

Nicogene:


Nicogene:
After an internal discussion, we will skip the superbuild step of release since this patch doesn't touch any of the repos covered by the superbuild

PR #449, opened 2020-08-25
Update 2020.08.yaml

Nicogene:
This PR updates the 2020.08.yaml for the upcoming release


traversaro (approved):
Thanks, let's see CI's opinion.

traversaro:
The 2020.08 tests are successful, while the Default/Stable are not working correctly due to https://github.com/robotology/robotology-superbuild/pull/451#issuecomment-680696161 , however as the problem is indipendent from this PR I think we can merge.

PR #451, opened 2020-08-25
Use YARP 3.4 as "stable" YARP branch

traversaro:
YARP 3.4.0 was released 25 days ago (see https://github.com/robotology/yarp/releases/tag/v3.4.0), and it will be included in the 2020.08 distro release, and the new gazebo-yarp-plugins 3.5 release will depend on it (see https://github.com/robotology/gazebo-yarp-plugins/pull/488#issuecomment-638153645).

For this reason, I guess it is time to start using YARP 3.4. as the "stable" branch. To avoid failures, for the time being we also use the "devel" branch of `gazebo-yarp-plugins` as stable branch, and we will switch back to use the `master` as soon as gazebo-yarp-plugins 3.5 is released.


diegoferigo (approved):

traversaro:
Tests are failing because icub-main devel devel has been merged in master, but icub-firmware-shared devel has not been merged in master. I can temporary set as "stable" branch of icub-firmware-shared the `devel` branch, but if the merge is happening soon probably it does not make sense to do so. @Nicogene

PR #459, opened 2020-09-02
Update README to point to the latest version of robotology-superbuild-dependencies-vcpkg

S-Dafarra:
For example, right now it points to the 0.3.0 version, which is not the latest. This would avoid us to update this README every time we have a new release.


traversaro (approved):

traversaro:
Good idea, thanks!

PR #458, opened 2020-09-01
GitHub Actions: install python3-numpy

traversaro:
This is now required since https://github.com/robotology/idyntree/pull/726 .

Without this fix, the unstable CI is failing with the error: 
~~~
2020-09-01T14:03:36.5315042Z CMake Error at /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
2020-09-01T14:03:36.5315306Z   Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found
2020-09-01T14:03:36.5315430Z   version "3.8.2")
2020-09-01T14:03:36.5315536Z Call Stack (most recent call first):
2020-09-01T14:03:36.5315828Z   /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393 (_FPHSA_FAILURE_MESSAGE)
2020-09-01T14:03:36.5316211Z   /usr/share/cmake-3.16/Modules/FindPython/Support.cmake:2214 (find_package_handle_standard_args)
2020-09-01T14:03:36.5316487Z   /usr/share/cmake-3.16/Modules/FindPython3.cmake:300 (include)
2020-09-01T14:03:36.5316611Z   bindings/python/CMakeLists.txt:3 (find_package)
~~~

PR #461, opened 2020-09-02
GitHub Actions: Switch Debian stable testing to use backports

traversaro:
This fixes the same problem as https://github.com/robotology/robotology-superbuild/pull/458 but for Debian Stable.


traversaro:
The only relevant job is  `[docker:Tags:Unstable@debian:buster-backports@Debug]` that passed fine. All other jobs are failing due to network flakes.

diegoferigo (approved):

Issue #463, opened 2020-09-03
Only use backslash when setting PATH and other related enviroment variables in Windows with Powershell or cmd


traversaro:
As reported by @S-Dafarra in a internal issue, the default installation instructions of the robotology-superbuild do not work correctly, because the PATH is populate by file paths with the forward slashes, while the PATH env variable apparently supports only backslash:
![powershell_wrong](https://user-images.githubusercontent.com/1857049/92084715-17fe3800-edc8-11ea-9e0c-b32b0ee4e502.png)

While this is a critical issue, it went unnoticed for the following reason: 
* I (and I guess some other users) also on Windows use the Git Bash, and source the environment using the `setup.sh` scripts, that due to the conversions in the PATH env variable between the Git enviroment and the Windows enviroment are not affected by this problem. 
* Most of the executables worked fine as they were in the same directory of the `.dll` they used, the only library that were affected by this were the YARP plugins in `bin/YARP`, `bin/iCub` or similar.

PR #464, opened 2020-09-03
For Powershell and Command Prompt, always use backslash for paths

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/463


S-Dafarra (approved):
Thanks!

Issue #469, opened 2020-09-10
Non-deterministic GitHub Action failure due to YARP file


traversaro:
Apparently, since yesterday some (apparently) non-deterministic  failures started to appear on GitHub Actions (see for example ). The error message changes from platform to platform, but it is tipically something like:
~~~
2020-09-10T02:04:31.5249461Z [  5%] Creating directories for 'YARP'
2020-09-10T02:04:31.5249915Z [  5%] Performing download step (git clone) for 'YARP'
2020-09-10T02:04:39.3121524Z Already on 'master'
2020-09-10T02:04:39.3123135Z M	extern/xmlrpcpp/xmlrpcpp/test/pngnow.png
2020-09-10T02:04:39.3124233Z Your branch is up to date with 'origin/master'.
2020-09-10T02:04:39.3483508Z [  6%] No patch step for 'YARP'
2020-09-10T02:04:39.3529603Z [  6%] Performing update step for 'YARP'
2020-09-10T02:04:40.1734192Z error: cannot rebase: You have unstaged changes.
2020-09-10T02:04:40.1736391Z error: Please commit or stash them.
2020-09-10T02:04:40.1757245Z fatal: No rebase in progress?
2020-09-10T02:04:40.2155098Z error: Your local changes to the following files would be overwritten by merge:
2020-09-10T02:04:40.2157406Z 	extern/xmlrpcpp/xmlrpcpp/test/pngnow.png
2020-09-10T02:04:40.2158192Z Please commit your changes or stash them before you merge.
2020-09-10T02:04:40.2158538Z Aborting
2020-09-10T02:04:40.2160483Z Index was not unstashed.
2020-09-10T02:04:40.2163306Z The stash entry is kept in case you need it again.
2020-09-10T02:04:40.2170819Z CMake Error at /home/runner/work/robotology-superbuild/robotology-superbuild/build/robotology/YARP/CMakeFiles/YCMTmp/YARP-gitupdate.cmake:105 (message):
2020-09-10T02:04:40.2171011Z   
2020-09-10T02:04:40.2171248Z 
2020-09-10T02:04:40.2171366Z   Failed to rebase in:
2020-09-10T02:04:40.2171702Z   '/home/runner/work/robotology-superbuild/robotology-superbuild/robotology/YARP/'.
2020-09-10T02:04:40.2171805Z 
2020-09-10T02:04:40.2171862Z 
2020-09-10T02:04:40.2171987Z   You will have to resolve the conflicts manually
[...]

PR #470, opened 2020-09-11
Temporary pin YARP stable and unstable version to a valid commit

traversaro:
Due to https://github.com/robotology/robotology-superbuild/issues/469, the robotology-superbuild on both Stable and Unstable branches is broken on Linux and macOs . As a temporary solution until https://github.com/robotology/yarp/pull/2356 is merged, let's pin YARP to some working commits.


traversaro:
The macOS failures are unrelated and due to https://github.com/osrf/homebrew-simulation/issues/1120 .

diegoferigo (approved):

PR #471, opened 2020-09-11
Restore regular YARP branches

traversaro:
As https://github.com/robotology/yarp/pull/2356 has been merged.

### DIFF
--- a/cmake/ProjectsTagsStable.cmake
+++ b/cmake/ProjectsTagsStable.cmake
@@ -10,5 +10,4 @@ set_tag(osqp_TAG v0.6.0)
 # Robotology projects
 set_tag(YCM_TAG ycm-0.11)
 set_tag(YARP_TAG yarp-3.4)
-set_tag(YARP_TAG 0328577b5f102da2875517898d8a6439ac9374d2)
 set_tag(yarp-matlab-bindings_TAG yarp-3.4)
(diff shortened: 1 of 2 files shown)


cmake/ProjectsTagsStable.cmake
```diff
@@ -9,6 +9,6 @@ set_tag(osqp_TAG v0.6.0)
 
 # Robotology projects
 set_tag(YCM_TAG ycm-0.11)
-# set_tag(YARP_TAG yarp-3.4)
+set_tag(YARP_TAG yarp-3.4)
 set_tag(YARP_TAG 0328577b5f102da2875517898d8a6439ac9374d2)
```
S-Dafarra, line 13: Probably this has to be removed

S-Dafarra (changes requested):

cmake/ProjectsTagsStable.cmake
traversaro: Thanks, done!

S-Dafarra (approved):

traversaro:
Apparently the change in https://github.com/robotology/yarp/pull/2356 still needs to be propagated to `master`, so we need to wait for that to merge this PR, because the `Stable` branches give an error as the default branch of YARP is `master`, and the error emerges before switching to `yarp-3.4`.

traversaro:
The automatic merging of `yarp-3.4` in `master` was broken by exactly the same problem described in https://github.com/robotology/yarp/pull/2356 . Now the merge has been done manually, so the PR is ready to be merged.

PR #474, opened 2020-09-24
Fix rtf setup

Nicogene:
The env vars setup scripts have not been updated since the rename from `rtf` to `robottestingframework`

Please review code


traversaro:
> The env vars setup scripts have not been updated since the rename from `rtf` to `robottestingframework`

For reference, this was done in https://github.com/robotology/yarp/pull/1998 .

traversaro (approved):

PR #482, opened 2020-10-10
BuildICUB: Remove options related to GTK2 GUIs

traversaro:
The option was removed in https://github.com/robotology/icub-main/pull/659 .


traversaro:
The macOS failure is a usual spurious and unrelated failure.

Nicogene (approved):

Issue #477, opened 2020-10-04
Document compilation against conda-forge provided dependencies


traversaro:
After https://github.com/conda-forge/staged-recipes/pull/12698, at least for Linux and Windows the dependencies of the robotology-superbuild should be all available in conda-forge. For this reason, it could make sense to document somewhere the procedure to compile the robotology-superbuild against the dependency provided by conda-forge .

traversaro:
The basic instructions are: 
~~~
# Install dependencies
conda install -c conda-forge ace asio boost eigen freeglut gsl glew glfw libxml2 libode libjpeg-turbo libmatio pkg-config opencv portaudio qt sdl sdl2 sqlite gazebo 
# Build 
# On *nix
export CMAKE_PREFIX_PATH=$CONDA_PREFIX
# On Windows
set CMAKE_PREFIX_PATH=%CONDA_PREFIX%
# Compile
git clone https://github.com/robotology/robotology-superbuild
cd robotology-superbuild
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release ..
cmake --build --config Release ..
~~~
However on Windows the compilation is failing with:
~~~
  YARP_priv_hmac.vcxproj -> C:\src\conda-ws\robotology-superbuild\build\robotology\YARP\lib\Release\YARP_priv_hmac.lib
     Creating library C:/src/conda-ws/robotology-superbuild/build/robotology/YARP/lib/Release/YARP_os.lib and object C:/src/conda-ws/robotology-superbuild/build/robotology/YARP/lib/Release/YARP_os.exp
Os.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) char const * __cdecl ACE::basename(char const *,char)" (__imp_?basename@ACE@@YAPEBDPEBDD@Z) referenced in function "void __cdecl yarp::o
s::setprogname(char const *)" (?setprogname@os@yarp@@YAXPEBD@Z) [C:\src\conda-ws\robotology-superbuild\build\robotology\YARP\src\libYARP_os\src\YARP_os.vcxproj]
Property.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) struct ACE_DIR * __cdecl ACE_OS::opendir_emulation(char const *)" (__imp_?opendir_emulation@ACE_OS@@YAPEAUACE_DIR@@PEBD@Z) referen
[...]

PR #484, opened 2020-10-11
CI: Add conda-forge based builds on Linux

traversaro:
See https://github.com/robotology/robotology-superbuild/issues/477 for more details. For now, only Linux is added, because some dependencies are still missing on macOS (freeglut) and Windows (ipopt).
 
In a nutshell conda is a package manager (originally for Python, but now quite language agnostic) that works on Linux, macOS and Windows. `conda-forge`  is a channel for the conda package manager that provides many dependency, in particular all the one that are required by the robotology-superbuild .

The use of conda is complementary to the compilation modes that we already use and are not substituted by conda, but its advantages are: 
* Updated dependencies: conda-forge tipically has relatively recent version of the dependencies, so if you want to use a recent vtk or pcl on an old distro such as Ubuntu 16.04, you can. This permits us to support older distributions even when the packages that install via apt are relatively old. 
* No need for root permissions: all the software installed by conda is installed and used in a user directory, so even if you are on a system in which you do not have root access (such as a shared workstation) you can still install all you required dependencies
* Use of binaries: conda distributes its packages as binaries, so even to download heavy dependencies such as OpenCV, PCL, Qt and Gazebo on Windows it just takes a few minutes, as opposed to hours necessary to compile them when using vcpkg. This is also useful when producing Docker images that require a recent version of PCL or VTK: installing them via conda takes a few seconds, and this would cut the time necessary to regenerate Docker images. 
* Reproducible enviroments: conda has built in support for installing exactly the same version of the packages you were using in the past, up to the patch version. This is quite important for reproducibility in scientific research. 
* On macOS and Windows, conda is also the package manager for which it is more easily possible to obtain working binaries of ROS1 and in the future of ROS2, thanks to the work of the [RoboStack project](https://github.com/RoboStack)

This PR does not add a build on Windows as ipopt binaries are still missing on that platform (see https://github.com/conda-forge/ipopt-feedstock/issues/1), and on macOS as freeglut binaries are missing (https://github.com/conda-forge/freeglut-feedstock/pull/24). The first practical use of this conda-based CI builds (once mac is added) is to quickly understand if any regression in the regular homebrew-based macOS CI builds is due to software regressions in the robotology-superbuild software, or in some enviroment failures (that is what happens most of the time).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,8 @@ jobs:
       matrix:
         build_type: [Release]
         # Windows is disabled due to the missing ipopt package
-        os: [ubuntu-latest, macOS-latest]
+        # macOS is disabled due to the missing freeglut package
+        os: [ubuntu-latest]
         cmake_generator: 
           - "Ninja"
         project_tags:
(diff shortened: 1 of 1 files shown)


traversaro:
The macOS build are failing due to the missing freeglut, that should be addressed in https://github.com/conda-forge/freeglut-feedstock/pull/24 .

traversaro:
> The first practical use of this conda-based CI builds (once mac is added) is to quickly understand if any regression in the regular homebrew-based macOS CI builds is due to software regressions in the robotology-superbuild software, or in some enviroment failures (that is what happens most of the time).

Just to make this point clear, the Homebrew-based CI is now failing on macOS.

diegoferigo (approved):
Cool! Nice to see a cross-platform system that can handle both source and binary installation. Furthermore, for researchers, being able to install the dependencies in a user-owned folder might provide a easier way to keep a clean system.

Issue #485, opened 2020-10-11
Release 2020.08.patch-01


traversaro:
Version to bump:
* YARP from v3.4.0 to v3.4.1
* YCM from v0.11.3 to v0.11.4 
* GazeboYARPPlugins from v3.5.0 to v3.5.1 (still needs to be released, see https://github.com/robotology/gazebo-yarp-plugins/pull/520) 
* icub-models from v1.17.0 to v1.17.1 (to fix the velocity control, see https://github.com/robotology/icub-models/pull/48)

traversaro:
@pattacini do you want to add some fixes for icub-main ? In that case it would be great to to a patch release of icub-main.

pattacini:
We may want to fix https://github.com/robotology/icub-main/issues/685.
Awaiting reply from @ale-git.

traversaro:
> 
> 
> We may want to fix [robotology/icub-main#685](https://github.com/robotology/icub-main/issues/685).
> Awaiting a reply from @ale-git.

Ack, even if (at least on my side) there is no urgency in fixing https://github.com/robotology/icub-main/issues/685 .

traversaro:
Yes, no problem! I just pinged you in the case you had some fix that you wanted to release, not because I wanted https://github.com/robotology/icub-main/issues/685 to be fixed anytime soon.

traversaro:
The release has been tagged at [`v2020.08.patch-01`](https://github.com/robotology/robotology-superbuild/tree/v2020.08.patch-01) and autogenerated Windows binaries are available at https://github.com/robotology/robotology-superbuild/releases/tag/v2020.08.patch-01 . Unless there is a specific problems, I would close the issue.

Issue #490, opened 2020-10-11
macOS Homebrew CI fails with openssl related problem


traversaro:
~~~
==> Upgrading openssl@1.1 1.1.1g -> 1.1.1h 
==> Downloading https://homebrew.bintray.com/bottles/openssl%401.1-1.1.1h.catalina.bottle.tar.gz
[...]

PR #492, opened 2020-10-11
Update versions in 2020.08.patch-01

traversaro:
See discussion in https://github.com/robotology/robotology-superbuild/issues/485 . Furthermore, I also set yarp-matlab-bindings to v3.4.0 that for an error was still pointing to `master` .


traversaro:
The homebrew CI failure can be ignored (see https://github.com/robotology/robotology-superbuild/issues/490).

pattacini (approved):

PR #497, opened 2020-10-17
GitHub Actions: Bump used vcpkg-robotology archive to v0.5.0

traversaro:
~The v0.5.0 is still a prerelease, if the compilation with these dependencies fails fixes will be deployed. Once the compilation with the new archive is working, the v0.5.0 release of robotology-superbuild-dependencies-vcpkg will be made official.~

The release is working fine, so it is now an official release: https://github.com/robotology/robotology-superbuild-dependencies-vcpkg/releases/tag/v0.5.0 .

> ~Furthermore, switch from using wget (that needs to be manually installed via `choco`) to using curl (that is already installed on Windows, see https://docs.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows).~

Obviously, trying to change two things at the same time always bite you back. I am reverting the wget/curl change for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -230,11 +230,11 @@ jobs:
       if: matrix.os == 'windows-2019'
       run: |
         # To avoid spending a huge time compiling vcpkg dependencies, we download an archive  that comes precompiled with all the ports that we need 
-        choco install -y unzip
+        choco install -y wget unzip
         # To avoid problems with non-relocatable packages, we unzip the archive exactly in the same directory
         # that has been used to create the pre-compiled archive
         cd C:/
-        curl https://github.com/robotology/robotology-superbuild-dependencies-vcpkg/releases/download/v0.5.0/vcpkg-robotology-with-gazebo.zip -o vcpkg-robotology-with-gazebo.zip
+        wget https://github.com/robotology/robotology-superbuild-dependencies-vcpkg/releases/download/v0.5.0/vcpkg-robotology-with-gazebo.zip
         unzip vcpkg-robotology-with-gazebo.zip -d C:/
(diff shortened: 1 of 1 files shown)


traversaro:
Windows CI is successful, PR ready for review.

.github/workflows/ci.yml
```diff
@@ -234,7 +234,7 @@ jobs:
         # To avoid problems with non-relocatable packages, we unzip the archive exactly in the same directory
         # that has been used to create the pre-compiled archive
         cd C:/
-        wget https://github.com/robotology/robotology-superbuild-dependencies-vcpkg/releases/download/v0.3.0/vcpkg-robotology-with-gazebo.zip
+        wget https://github.com/robotology/robotology-superbuild-dependencies-vcpkg/releases/download/v0.5.0/vcpkg-robotology-with-gazebo.zip
```
S-Dafarra, line 237: Do you prefer to control the version, or we could always download the latest?

traversaro: I think it is important to control the version. I prefer to open a PR for any new release of `robotology-superbuild-dependencies-vcpkg` to explicitly check that the new release works fine, rather then do a release of  `robotology-superbuild-dependencies-vcpkg` that may break the robotology-superbuild CI, and then be force to a rollback PR to robotology-superbuild to fix the problem. As we use the latest archive in the docs, if the latest does not work correctly it is still a problem that needs to address fast, but I prefer not to break CI that would basically block any other PR from being merged without forcing.

S-Dafarra: Makes sense!

S-Dafarra (approved):

Issue #500, opened 2020-10-19
Prepare release 2020.11 


traversaro:
As we are getting used to, it is time to start preparing the 2020.11 release.

traversaro:
The release process will follow the workflow documented in the Developer's FAQs : https://github.com/robotology/robotology-superbuild/blob/master/doc/developers-faqs.md#how-to-do-a-new-release .

traversaro:
Hi @arrenglover @francodipietro, given that https://github.com/robotology/robotology-superbuild/pull/502 was just merged, it would be nice if you could tag a version of the event-driven so that it can be included in the 2020.11 repo. We typically use the v1.2.3 style for tags, but also any other style is perfectly ok, thanks!

S-Dafarra:
I just noticed that after https://github.com/robotology/idyntree/pull/751 gets merged, the ``UnicyclePlanner`` would not compile. Working on a release for it.

traversaro:
> 
> 
> I just noticed that after [robotology/idyntree#751](https://github.com/robotology/idyntree/pull/751) gets merged, the `UnicyclePlanner` would not compile. Working on a release for it.

Probably the `UnicyclePlanner`  was using some deprecated iDynTree function? Or there were some problems in https://github.com/robotology/idyntree/pull/751 ?

S-Dafarra:
> Probably the `UnicyclePlanner` was using some deprecated iDynTree function? Or there were some problems in [robotology/idyntree#751](https://github.com/robotology/idyntree/pull/751) ?

I forgot some deprecated methods in there :grin:

S-Dafarra:
> I just noticed that after [robotology/idyntree#751](https://github.com/robotology/idyntree/pull/751) gets merged, the `UnicyclePlanner` would not compile. Working on a release for it.

Done, #507

traversaro:
@prashanthr05 @HosameldinMohamed if you do not see any blocking problem, it would be also useful to merge devel in master and do a new release of https://github.com/robotology/whole-body-estimators .

prashanthr05:
> @prashanthr05 @HosameldinMohamed if you do not see any blocking problem, it would be also useful to merge devel in master and do a new release of https://github.com/robotology/whole-body-estimators .

Yes, we shall do that.

prashanthr05:
[...]

PR #499, opened 2020-10-19
Add portaudio dependency and enable audio-related devices in YARP

traversaro:
This PR add the dependency on portaudio on Linux, macOS and Windows, and enable by default the compilation of the YARP devices related to audio. Fix https://github.com/robotology/robotology-superbuild/issues/284 .

PR #501, opened 2020-10-20
Add initial files for the 2020.11 release

traversaro:
First step for https://github.com/robotology/robotology-superbuild/issues/500 .
This PR adds a `2020.11.yaml` file, bump it the YCM version to an unreleased commit from the 0.12 branch. Furthermore, the 2020.11 release is added to the CI, while the older relase are removed.
The rationale is that old releases will be broken by the enabling of portaudio dependencies, see  https://github.com/robotology/robotology-superbuild/pull/499#issuecomment-712121774  . 
For use old releases, it is still possible to checkout the relative tag, as discussed in the release notes (see for example https://github.com/robotology/robotology-superbuild/releases/tag/v2020.08).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
       matrix:
         build_type: [Release]
         os: [ubuntu-latest, macOS-latest, windows-2019]
-        project_tags: [Default, Unstable, Release202005feat01, Release202008patch01]
+        project_tags: [Default, Unstable, Release202011]
         include:
           - os: ubuntu-latest
             build_type: Debug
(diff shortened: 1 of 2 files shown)


traversaro:
Please ignore failures in Debian Sid, they are related to https://github.com/robotology/robotology-superbuild/pull/498 .

.github/workflows/ci.yml
```diff
@@ -142,7 +142,7 @@ jobs:
       matrix:
         build_type: [Release]
         os: [ubuntu-latest, macOS-latest, windows-2019]
-        project_tags: [Default, Unstable, Release202005feat01, Release202008patch01]
+        project_tags: [Default, Unstable, Release202011]
```
Nicogene, line 145: Did you reduce the releases tested from 2 to 1 for build time issues?

traversaro: No, we remove the old releases as they will break when we add the portaudio dependencies, see https://github.com/robotology/robotology-superbuild/pull/499#issuecomment-712121774 for more details.

Nicogene (approved):

PR #504, opened 2020-10-20
GitHub Actions: Enable conda-forge CI on macOS

traversaro:


traversaro:
Both conda and mamba fail as the package cannot be installed on macOS. The conda error is quite long, while the mamba one is succinct: 
> Problem: package freeimage-3.18.0-h8ceafe6_3 requires openexr >=2.5.1,<2.6.0a0, but none of the providers can be installed

I am not an expert but a lot of problems seems to be related to Gazebo that still depends on a lot of old libraries, so probably a new build of Gazebo could help.

traversaro:
Closing for now, we can re-open when the gazebo package is updated.

PR #509, opened 2020-10-29
GitHub Actions: Fix Homebrew problem

traversaro:
A lot of repo are failing on `brew upgrade` with error: 
~~~
==> Pouring python@3.9-3.9.0_1.catalina.bottle.tar.gz
Error: The `brew link` step did not complete successfully
Warning: These files were overwritten during `brew link` step:
The formula built, but is not symlinked into /usr/local
Could not symlink lib/pkgconfig/python3-embed.pc
Target /usr/local/lib/pkgconfig/python3-embed.pc
is a symlink belonging to python@3.8. You can unlink it:
  brew unlink python@3.8
~~~

Let's try to fix those. As first step, let's try to remove the workaround added for https://github.com/actions/virtual-environments/issues/1811 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -211,6 +211,8 @@ jobs:
         cmake --version
         brew cask install xquartz
         # Update homebrew 
+        # Workaround for https://github.com/actions/virtual-environments/issues/1811#issuecomment-718475660
+        brew unlink python@3.8
         brew update
         brew upgrade
         # Core dependencies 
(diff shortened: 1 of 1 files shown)


traversaro:
The problem is probably due to https://github.com/Homebrew/homebrew-core/issues/63619#issuecomment-718404594 . 
cc @Nicogene

traversaro:
The error is the same even after removing the https://github.com/actions/virtual-environments/issues/1811 workarounds

traversaro:
The workaround of adding `brew unlink python@3.8` seems to be effective (see https://github.com/robotology/robotology-superbuild/pull/509/checks?check_run_id=1325474151).

Nicogene (approved):

PR #513, opened 2020-11-15
GitHub Actions: Enable conda-forge CI on macOS

traversaro:
Modern version of https://github.com/robotology/robotology-superbuild/pull/504 . Now the gazebo build has been refreshed, so everything should work fine. 

For more details on what conda and conda-forge are, and why it useful to have a CI with them, please see https://github.com/robotology/robotology-superbuild/pull/484#issue-501112882 . For the specific case of macOS, having an build that is not using Homebrew should permit us to distinguish the quite frequent Homebrew failures from the actual macOS compilation problems (that should create a failure in both conda and Homebrew macOS jobs).

PR #514, opened 2020-11-15
 GitHub Actions: Enable conda-forge CI on Windows 

traversaro:
Based on https://github.com/robotology/robotology-superbuild/pull/513 for simplicity. Now that ipopt is available on Windows thanks to https://github.com/conda-forge/ipopt-feedstock/pull/47, at least compilation should work fine also on Windows, but let's see.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
       shell: bash -l {0}
       run: |
         # Additional dependencies only useful on Windows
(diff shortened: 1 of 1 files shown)


traversaro:
There are still some problems, mainly related to the fact that there is no `vs2017_win64`-like package for Visual Studio 2019, and GitHub Actions only supports  Visual Studio 2019.

traversaro:
This was useful for understanding the problems that even end users could face. Let's close for now.

traversaro:
-

PR #523, opened 2020-11-19
Update whole-body-controllers and whole-body-estimators tags for release 2020.11

traversaro:
The new release of `whole-body-estimators` was reported by @prashanthr05  in https://github.com/robotology/robotology-superbuild/issues/500#issuecomment-724180794 . 

For the `whole-body-controllers`, with @CarlottaSartore we noticed that the 2020.08-patch01 contained a rather old version of `whole-body-controllers`, so we bumped it. fyi @gabrielenava

### DIFF
--- a/releases/2020.11.yaml
+++ b/releases/2020.11.yaml
@@ -90,11 +90,11 @@ repositories:
   whole-body-controllers:
     type: git
     url: https://github.com/robotology/whole-body-controllers.git
-    version: v2.0
+    version: v2.5
   whole-body-estimators:
     type: git
     url: https://github.com/robotology/whole-body-estimators.git
-    version: v0.2.1
+    version: v0.3.0
   walking-teleoperation:
     type: git
     url: https://github.com/robotology/walking-teleoperation.git


gabrielenava:
ok, but what is exactly 2020.08-patch01?

traversaro:
> ok, but what is exactly 2020.08-patch01?

The release policy is explained in https://github.com/robotology/QA/issues/377 . Basically, every three months we do a new (time based) release, that is basically just a collection of release for all the superbuild subprojects, that has the name vYYYY.MM . Those offer a non-changed version for users that don't want to have a rolling release process typical of the default settings of superbuild, for example joint labs. Sometimes, due to bugs it may be necessary to do small updates to some projects after a time-based release, and those release are called vYYYY.MM-patchNN . So the latest release is the v2020.08-patch01 as it was a patch update over v2020.08, and the next release will be 2020.11, and that is the release for which I am including wbc 2.5 .

traversaro:
> 
> 
> > ok, but what is exactly 2020.08-patch01?
> 
> The release policy is explained in [robotology/QA#377](https://github.com/robotology/QA/issues/377) . Basically, every three months we do a new (time based) release, that is basically just a collection of release for all the superbuild subprojects, that has the name vYYYY.MM . Those offer a non-changed version for users that don't want to have a rolling release process typical of the default settings of superbuild, for example joint labs. Sometimes, due to bugs it may be necessary to do small updates to some projects after a time-based release, and those release are called vYYYY.MM-patchNN . So the latest release is the v2020.08-patch01 as it was a patch update over v2020.08, and the next release will be 2020.11 (#500), and that is the release for which I am including wbc 2.5 .

A more permanent description of the distro policy is present in https://icub-tech-iit.github.io/documentation/sw_versioning_table .

prashanthr05 (approved):

Issue #525, opened 2020-11-21
YCM Boostrap broken in recent Xcode and/or CMake (i.e. on GitHub Actions macos-* image)


traversaro:
This will be fixed by https://github.com/robotology/ycm/pull/364 .

PR #527, opened 2020-11-22
CI: Cleanup CI scripts

traversaro:
* Be explicit and avoid the use of `<os>-latest` images 
* Add Ubuntu-20.04 job without Docker
* Change checks on operating system to use contains instead of exact match of the string


diegoferigo (approved):

GiulioRomualdi (approved):

traversaro:
The failing jobs are due to https://github.com/robotology/robotology-superbuild/issues/525, so I think we can ignore them.

PR #528, opened 2020-11-22
Add FW versions

pattacini:
This PR adds up the packages related to FW in order to be able to produce the corresponding SW Versioning Table.


traversaro (approved):
Thanks @pattacini !

PR #521, opened 2020-11-18
CI: Add check if yarp devices are correctly listed

traversaro:
This should permit to debug problems such as https://github.com/robotology/wearables/issues/91 .

PR #533, opened 2020-11-24
Update wearables tag for release 2020.11

lrapetti:
After release of wearables `v1.1.0`, I am updating the tag for Release 2020.11.

cc @Yeshasvitvs

### DIFF
--- a/releases/2020.11.yaml
+++ b/releases/2020.11.yaml
@@ -106,7 +106,7 @@ repositories:
   wearables:
     type: git
     url: https://github.com/robotology/wearables.git
-    version: v1.0.0
+    version: v1.1.0
   human-dynamics-estimation:
     type: git
     url: https://github.com/robotology/human-dynamics-estimation.git


lrapetti:
Note that, we have been informed of problems with some Windows installation (https://github.com/robotology/wearables/issues/91). We are investigating it and in case we will soon provide a patch for that.

traversaro:
> 
> 
> Note that, we have been informed of problems with some Windows installation ([robotology/wearables#91](https://github.com/robotology/wearables/issues/91)). We are investigating it and in case we will soon provide a patch for that.

Indeed, as the problem is there also in the 1.0.0 release (see the `[Tags:Release202011@windows-2019@Release]` job in https://github.com/robotology/robotology-superbuild/pull/521/checks?check_run_id=1427138013 ) I do not think it is a problem to bump the version.

lrapetti:
> Indeed, as the problem is there also in the 1.0.0 release

Then I think we can proceed

traversaro (approved):

traversaro:
Thanks @lrapetti , it just fails on jobs that are failing due to https://github.com/robotology/robotology-superbuild/issues/525, let's merge anyway.

PR #534, opened 2020-11-24
Drop compatibility with Xcode 12

traversaro:
For the motivations described in https://github.com/robotology/ycm/issues/368, supporting Xcode 12 is quite complicated at the moment, so we just remove it from the CI jobs and we will document that we don't support it.


drdanz:
For reference, I'm planning to release 0.12.1 in a couple of days...

traversaro:
> 
> 
> For reference, I'm planning to release 0.12.1 in a couple of days...

Exactly, this is for starting gaining time and test the 0.12.1 on all the CIs jobs, and catch early any possible problem.

lrapetti (approved):

traversaro:
> 
> 
> > For reference, I'm planning to release 0.12.1 in a couple of days...
> 
> Exactly, this is for starting gaining time and test the 0.12.1 on all the CIs jobs, and catch early any possible problem.

Indeed, CI is still not working.

drdanz:
I'll stand by for the release then :laughing:

traversaro:
> 
> 
> > > For reference, I'm planning to release 0.12.1 in a couple of days...
> > 
> > 
> > Exactly, this is for starting gaining time and test the 0.12.1 on all the CIs jobs, and catch early any possible problem.
> 
> Indeed, CI is still not working.

There were still some problem, hopefully fixed by https://github.com/robotology/ycm/pull/367 . As a test I included the commit of the PR in the release file for now.

traversaro:
The version of YCM set in the .yaml was ignored, because there was a bug in the .yaml file because YCM was referred with the repo name `ycm`, instead of which the CMake package name `YCM` . This is not the first time it happens, so it is better to open an issue on how to to fix this in the long term.

traversaro:
Ok, now  a new error: 
~~~
-- Generating done
  The custom command generating

    /Users/runner/work/robotology-superbuild/robotology-superbuild/build/robotology/YCM/CMakeFiles/YCMStamp/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)/YCM-mkdir

  is attached to multiple targets:

    YCM-update
    YCM

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".


CMake Generate step failed.  Build files cannot be regenerated correctly.
~~~

Issue #505, opened 2020-10-21
Document how to consume releases 


traversaro:
The workflow to consume releases is just to checkout the release tag (so `v2020.08`, `v2020.08-patch01`, etc etc) and then just proceed to configure the superbuild as usual, as for those tags the default value of `ROBOTOLOGY_PROJECT_TAGS` is `Custom` and the default value of `ROBOTOLOGY_PROJECT_TAGS_CUSTOM_FILE` is pointing to the correct file. 

This is documented at the top of the release notes for each release, but it is not mentioned at all in the main README, so it may be hard for users to get this info. 

cc @CarlottaSartore

PR #535, opened 2020-11-24
Restructure README moving CMake options documentation in separate file

traversaro:
Over the years the README of the superbuild has grown complex. To make it more readable, let's move the CMake options-related documentation in another page, and just keep installation instructions in the main README. This change is in view of fixing https://github.com/robotology/robotology-superbuild/issues/505 .


Nicogene (approved):
I love it!

PR #544, opened 2020-11-28
Enable cppad on Windows

traversaro:
In particular, check if https://github.com/coin-or/CppAD/issues/60 is fixed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,18 +169,6 @@ if(ROBOTOLOGY_ENABLE_DYNAMICS)
   endif()
 endif()
 
-if(ROBOTOLOGY_ENABLE_DYNAMICS_FULL_DEPS)
-  find_or_build_package(manif)
-  find_or_build_package(qhull)
-  find_or_build_package(casadi)
-
-  # cppad is currently disabled on windows
-  # https://github.com/robotology/robotology-superbuild-dependencies-vcpkg/pull/37
-  if(NOT WIN32)
-    find_or_build_package(CppAD)
-  endif()
-endif()
-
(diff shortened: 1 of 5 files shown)


traversaro:
There is a unrelated conda/macOS failure (it should be fixed by https://github.com/conda-forge/gazebo-feedstock/pull/38) but other then that the fix seems to be working fine, if we update cppad to the latest version.

CMakeLists.txt
```diff
@@ -169,18 +169,6 @@ if(ROBOTOLOGY_ENABLE_DYNAMICS)
   endif()
 endif()
 
-if(ROBOTOLOGY_ENABLE_DYNAMICS_FULL_DEPS)
-  find_or_build_package(manif)
-  find_or_build_package(qhull)
-  find_or_build_package(casadi)
-
```
GiulioRomualdi, line 176: This part shouldn't be removed right?

traversaro: Sorry, I forgot to describe this in the PR. As we are including these as dependencies of blf, there is no need to also duplicate these in the main CMakeLists.txt . I realized this when I had two modified in two places to enable cppad on Windows.

GiulioRomualdi:
On `macOS` s failing but since it does not depend on this PR I think you can merge it, right?

traversaro:
> 
> 
> On `macOS` s failing but since it does not depend on this PR I think you can merge it, right?

Yes, macOS/Conda failed but it was fixed by https://github.com/conda-forge/gazebo-feedstock/pull/38 .

GiulioRomualdi (approved):

PR #548, opened 2020-12-05
Cleanup main CMakeLists.txt file 

traversaro:
Create two new files:
* `cmake\RobotologySuperbuildOptions.cmake` that contain the options of the superbuild
* `cmake\RobotologySuperbuildLogic.cmake` that contain the logic of which projects need to be built (and their cmake options) depending on options enabled.

This should make the main `CMakeLists.txt` file more readable and mantainable.


traversaro:
The failure are not directly related to the PR, but due to https://github.com/robotology/robotology-superbuild/issues/549  and should be solved as I now created again the devel branch.

Nicogene (approved):
+1

Issue #552, opened 2020-12-09
Regression in conda-forge Linux CI


traversaro:
From https://github.com/robotology/robotology-superbuild/actions . 

Two days ago everything was working fine: 
* https://github.com/robotology/robotology-superbuild/runs/1514709988?check_suite_focus=true 

Yesterday instead something went wrong, and went wrong also in https://github.com/robotology/robotology-superbuild/pull/532 : https://github.com/robotology/robotology-superbuild/runs/1521342051?check_suite_focus=true 

The error is: 
~~~
Looking for: ['ace', 'asio', 'boost', 'eigen', 'gazebo', 'glew', 'glfw', 'gsl', 'ipopt', 'libjpeg-turbo', 'libmatio', 'libode', 'libxml2', 'opencv', 'pkg-config', 'portaudio', 'qt', 'sdl', 'sdl2', 'sqlite', 'tinyxml']

Encountered problems while solving.
Problem: package gazebo-11.0.0-h398b6f7_9 requires qt >=5.12.5,<5.13.0a0, but none of the providers can be installed
~~~

I briefly tried to replicate this locally, and it seems that instead conda is working fine, I spawned https://github.com/robotology/robotology-superbuild/pull/551 to check if this happens also on the CI.

traversaro:
Cool, switching to conda as done in https://github.com/robotology/robotology-superbuild/pull/551 fixes the problem on Linux, and it creates a new problem on macOS: https://github.com/robotology/robotology-superbuild/pull/551/checks?check_run_id=1527274310 .

traversaro:
Indeed, the problem is in mamba. 
~~~
mamba install -c conda-forge cmake compilers make ninja pkg-config ace asio boost eigen gazebo glew glfw gsl ipopt libjpeg-turbo libmatio libode libxml2 opencv pkg-config portaudio qt sdl sdl2 sqlite tinyxml
~~~
fails with the previous error (see ci job https://github.com/robotology/robotology-superbuild/runs/1529533907), while :

~~~
 conda install -c conda-forge cmake compilers make ninja pkg-config ace asio boost eigen gazebo glew glfw gsl ipopt libjpeg-turbo libmatio libode libxml2 opencv pkg-config portaudio qt sdl sdl2 sqlite tinyxml
~~~
works fine (see ci job https://github.com/robotology/robotology-superbuild/runs/1530469671).

[...]

PR #553, opened 2020-12-11
GitHub Actions: Install mamba as documented in setup-miniconda

traversaro:
Switch to install mamba and conda-forge channels as documented in https://github.com/conda-incubator/setup-miniconda/tree/v2.0.1#example-6-mamba , instead of using an ad-hoc procedure, that probably is creating problem, see https://github.com/robotology/robotology-superbuild/issues/552 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,17 +42,17 @@ jobs:
 
     - uses: conda-incubator/setup-miniconda@v2
       with:
-        auto-update-conda: true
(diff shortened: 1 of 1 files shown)


traversaro:
Homebrew failures will be fixed by  https://github.com/actions/virtual-environments/issues/2248, while this indeed fixes https://github.com/robotology/robotology-superbuild/issues/552 .

diegoferigo (approved):

Issue #543, opened 2020-11-27
Remove icub-gazebo-wholebody? 


traversaro:
I am not sure if someone is still using the models and worlds in https://github.com/robotology/icub-gazebo-wholebody . If someone does, please comment here. If no one will reply, we should consider removing the repo. 

cc @robotology/iit-dynamic-interaction-control @gabrielenava

gabrielenava:
I don't use them anymore, maybe @Yeshasvitvs for the iCub standing up from the chair?

Yeshasvitvs:
I am not using then repo models actively. @lrapetti please confirm if you are using this models for pRRI ?

lrapetti:
> @lrapetti please confirm if you are using this models for pRRI ?

I am not using those models, so for me we can proceed removing it.

PR #555, opened 2020-12-13
Remove icub-gazebo-wholebody

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/543


Yeshasvitvs (approved):

Yeshasvitvs:
@traversaro is there a changelog for superbuild ?

traversaro:
> 
> 
> @traversaro is there a changelog for superbuild ?

Not at the moment, even if we can think of adding it in the future.

Issue #568, opened 2021-01-04
Update homebrew docs to install qt@5 instead of qt (that will soon point to qt6)


traversaro:
Once https://github.com/Homebrew/homebrew-core/pull/68241/files lands.

PR #572, opened 2021-01-05
Update brew documentation and CI scripts to refer to qt@5

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/568, even if actually we never migrate from `qt5` to `qt`, so we will not have problems if `qt` will install `qt6` .


traversaro:
CI will fail for https://github.com/robotology/robotology-superbuild/issues/568 , but it is probably a conflict between GitHub Actions and Homebrew, not an intrisic problem of Homebrew or of this change.

kouroshD (approved):

Issue #571, opened 2021-01-05
macOS Homebrew CI is failing


traversaro:
Similar to https://github.com/robotology/robotology-superbuild/issues/565, the CI is now failing with:
~~~
2021-01-05T15:48:44.8228350Z [34m==>[0m [1mPouring gcc@9-9.3.0_1.catalina.bottle.tar.gz[0m
2021-01-05T15:48:50.8478390Z [31mError:[0m The `brew link` step did not complete successfully
2021-01-05T15:48:50.8479120Z The formula built, but is not symlinked into /usr/local
2021-01-05T15:48:50.8483440Z Could not symlink bin/c++-9
2021-01-05T15:48:50.8484230Z Target /usr/local/bin/c++-9
2021-01-05T15:48:50.8485440Z is a symlink belonging to gcc@9. You can unlink it:
2021-01-05T15:48:50.8485930Z   brew unlink gcc@9
2021-01-05T15:48:50.8486170Z 
2021-01-05T15:48:50.8486600Z To force the link and overwrite all conflicting files:
2021-01-05T15:48:50.8487530Z   brew link --overwrite gcc@9
2021-01-05T15:48:50.8487820Z 
2021-01-05T15:48:50.8488190Z To list all files that would be deleted:
2021-01-05T15:48:50.8489020Z   brew link --overwrite --dry-run gcc@9
2021-01-05T15:48:50.8489340Z 
2021-01-05T15:48:50.8489710Z Possible conflicting files are:
2021-01-05T15:48:50.8490630Z /usr/local/bin/c++-9 -> /usr/local/Cellar/gcc@9/9.3.0/bin/c++-9
2021-01-05T15:48:50.8493490Z /usr/local/bin/cpp-9 -> /usr/local/Cellar/gcc@9/9.3.0/bin/cpp-9
2021-01-05T15:48:50.8497350Z /usr/local/bin/g++-9 -> /usr/local/Cellar/gcc@9/9.3.0/bin/g++-9
2021-01-05T15:48:50.8500770Z /usr/local/bin/gcc-9 -> /usr/local/Cellar/gcc@9/9.3.0/bin/gcc-9
2021-01-05T15:48:50.8504730Z /usr/local/bin/gcc-ar-9 -> /usr/local/Cellar/gcc@9/9.3.0/bin/gcc-ar-9
[...]

PR #573, opened 2021-01-07
Fix macOS CI

traversaro:
Add a workaround for https://github.com/robotology/robotology-superbuild/issues/571 .


traversaro:
The fix is working, the only macOS failure is spurious.

Nicogene (approved):

PR #582, opened 2021-01-12
Fix bipedal-locomotion-framework version for Stable branches

traversaro:
blf master now depends on iDynTree devel, so until the next iDynTree release let's use an old commit of bipedal-locomotion-framework .


GiulioRomualdi (approved):
Thank you @traversaro!!

PR #596, opened 2021-01-19
Automatize the process of bumping the released version of superbuild subprojects

traversaro:
This PR changes the workflow to update the releases of the subprojects. Until this PR, maintainers had to manually bump the version of the subprojects that got a new release in the currently tested `<yyyy.mm.yaml>` file. This procedure was typically error prone, as it is was common to forget to do this manual bump. 

For this reason, the procedure as been changed as in the following:
* The superbuild now contains a `releases/latest.releases.yaml` file that is meant to contain the latest release of each subproject of the superbuild.
* This file is automatically updated by the `update-latest-releases.yml` GitHub action, that periodically checks the default branches of the repo (the one used if `ROBOTOLOGY_PROJECT_TAGS` is set to `Stable`) and extract the latest tag done on that branch. The action can also be run manually using the [`workflow_dispatch` event](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). The action automatically creates a Pull Request, so that the compilation of the new proposed tags is tested before they are merged. An example of such automatically generated pull request is https://github.com/traversaro/robotology-superbuild/pull/5 .
* If you have any project for which you want to manually manage the release used in the `releases/latest.releases.yaml` file,
you can disable the automatical update of the tags by adding its CMake name in the `projects_to_skip` array in the `scripts/robotologyUpdateLatestReleases.sh` script.
* When a new release needs to be prepared, the  `releases/latest.releases.yaml` file is copied to a `releases/<yyyy.mm>.yaml` file, and then the release workflow continues as usual.

The main good news for **project and components mantainers** is that **you do not need anymore to bump the version in the yaml file**, but this will be handled automatically by the `update-latest-releases.yml` GitHub Action. The only thing that you need to do is to to a tag on your stable branch.

@prashanthr05 @gabrielenava @S-Dafarra @lrapetti @Yeshasvitvs @arrenglover @GiulioRomualdi @kouroshD


Nicogene (approved):
Super cool!
If I understood correctly it would check once per week if there are new tags right?
What happens if no new tag has been added? Does the action do nothing? Or it opens the PR with no changes?(No idea if it is something possible)

traversaro:
>  If I understood correctly it would check once per week if there are new tags right?

Yes, either that or by running the action manually with workflow_dispatch . Given that this action just generate a PR, that in any case will not be merged unless the maintainer merge, something cool that we could even do is to activate the action via a comment in any issue, for example activate it if someone writes "@robotology-bot, please update latest releases", so that any user can activate it and run the CI for the updated version list even if the mantainer is not there, but let's do one thing at the time. 

>  What happens if no new tag has been added? Does the action do nothing? Or it opens the PR with no changes?(No idea if it is something possible)

To be honest I did not tried, but the documentation seems to be clear (see https://github.com/marketplace/actions/create-pull-request) on this:
> If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently.

Issue #597, opened 2021-01-20
Reduce macOS builds in CI


traversaro:
macOS builds have long queues and are slow. It is a good idea to limit them to the essentials, my proposal is to limit them to:
* Only 10.15 build for conda, Stable/Unstable branches 
* Only 11.0 build for Homebrew, Stable/Unstable and releases branches

traversaro:
In two weeks we will remove Homebrew builds from the CI (see https://github.com/robotology/robotology-superbuild/issues/842), so I think we can close this issue.

PR #600, opened 2021-01-22
Remove macOS 11.0 jobs

traversaro:
First step in the direction of https://github.com/robotology/robotology-superbuild/issues/597 . 

Having a nice subset of the combinations of macOS builds would be better, but apparently the logic with which GitHub Actions deals with the combinations has changed so doing that would require looking into that. As we never had problems just on 11.0 that were not on 10.15, I think we can first doing that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,7 +166,7 @@ jobs:
       fail-fast: false
       matrix:
         build_type: [Release]
-        os: [ubuntu-18.04, ubuntu-20.04, macos-10.15, macos-11.0, windows-2019]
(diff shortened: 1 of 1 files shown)


Nicogene (commented):
I probably missed some steps, but why we don' remove 10.15 instead?

traversaro:
> 
> 
> I probably missed some steps, but why we don' remove 10.15 instead?

It is safer to test with older versions, so you catch if some project accidentally uses an API or compiled feature that was introduced in 11.0 but it is not available in 10.15. Furthermore, the 11.0 are the machines for which the GitHub Actions teams is having some problems (see https://github.com/actions/virtual-environments/issues/2486) so I think it make sense to stick to 10.15 .

PR #606, opened 2021-01-30
Add dependency of blocktestcore on YCM

traversaro:
The YCM dependency was added in https://github.com/robotology/blocktest/commit/dce9197e5b58adf37697013a9736fc3cb0a246e6 but was never propagated here. 

We never realized this because by chance YCM always got compiled first and blocktestcore was lucky enough to find it. However, I noticed that the dependency was missing by experimenting with conda package generation in https://github.com/traversaro/robotology-superbuild/runs/1797950461 . While generating conda packages, each package is compiled in a sandbox where only the explicitly specified dependencies are available, and so there I noticed that blockfactory was not compiling without YCM.

### DIFF
--- a/cmake/Buildblocktestcore.cmake
+++ b/cmake/Buildblocktestcore.cmake
@@ -4,10 +4,13 @@
 
 include(YCMEPHelper)
 
+find_or_build_package(YCM QUIET)
+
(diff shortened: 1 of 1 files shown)


Nicogene (approved):
Nice catch, thanks!

PR #619, opened 2021-02-09
Set GAZEBO_YARP_PLUGINS_HAS_OPENCV to ON in gazebo-yarp-plugins compilation

traversaro:
I was looking into some Dockerfiles used for https://robot-bazaar.iit.it/homepage, and I noticed that there were a few workaround for enabling the GAZEBO_YARP_PLUGINS_HAS_OPENCV  option in gazebo-yarp-plugins compilation. However, I don't remember nor see any reason why that options should be disabled: OpenCV is a dependency of the `Core` profile of the robotology-superbuild, and so enabling  this option just take a bit of space more.


traversaro:
cc @vtikha @vvasco (please tag anyone else that could be interested)

vtikha (approved):
Thank you @traversaro , this makes things much simpler and quicker

Issue #605, opened 2021-01-30
Raise an error if the Update Latest Releases is not able to update the tag


traversaro:
Currently for some projects  the Update Latest Releases is not able to update their tag, printing instead the message (see https://github.com/robotology/robotology-superbuild/runs/1755632387?check_suite_focus=true ):
~~~
fatal: No names found, cannot describe anything.
~~~

This error should create a failure in the action, to explicitly catch if some project have some problems with releases (for example, I did not noticed myself that https://github.com/dic-iit/matio-cpp did not have release when we added it). For specific case in which we don't want to update the version, the project should be explicitly skipped by adding it in the list of skipped projects, see https://github.com/robotology/robotology-superbuild/blob/master/scripts/robotologyUpdateLatestReleases.sh#L20 .

Furthermore, we should probably also make the action run (even without opening any PR) on all PRs, to catch if a new project is introduced without updating the `latest.releases.yaml` file.

PR #627, opened 2021-02-21
robotologyUpdateLatestReleases: add message when checking for a new version of a package

traversaro:
This simplifies understanding to which package an error is related to. This is related to https://github.com/robotology/robotology-superbuild/issues/605, even if it does not address it in full.

### DIFF
--- a/scripts/robotologyUpdateLatestReleases.sh
+++ b/scripts/robotologyUpdateLatestReleases.sh
@@ -59,6 +59,8 @@ updateLatestRelease () {
     is_contained=$?
     if [ $is_contained == 1 ]
     then
+       # Printing package name to simplify debugging for errors
+        echo "Checking if a new release of ${package_name} is available"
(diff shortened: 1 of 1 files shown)


traversaro:
I tried the new script in https://github.com/robotology/robotology-superbuild/runs/1945527008?check_suite_focus=true and it seems to work as intended, in particular I know know that the two errors are related to `bipedal-locomotion-framework` and `yarp-device-ovrheadset`, that should both be solved soon:
* https://github.com/dic-iit/bipedal-locomotion-framework/issues/197
* https://github.com/robotology/yarp-device-ovrheadset/issues/3

traversaro:
The CI failures are unrelated and are going to be fixed by https://github.com/conda-forge/libignition-msgs1-feedstock/pull/29 , so I think we can merge this PR.

GiulioRomualdi (approved):

Issue #637, opened 2021-02-25
Windows installer generation fails due to be running out of space in GitHub Actions image


traversaro:
The recently pre-released v2021.02 tag is failing to create the Windows installer due to running out of space in the GitHub Actions image, example errors:
~~~
2021-02-25T11:24:13.9762981Z        Creating library D:/a/robotology-superbuild/robotology-superbuild/build/src/YARP/lib/qt5/qml/Release/robotology/yarp/scope/QtYARPScopePlugin.lib and object D:/a/robotology-superbuild/robotology-superbuild/build/src/YARP/lib/qt5/qml/Release/robotology/yarp/scope/QtYARPScopePlugin.exp
2021-02-25T11:24:13.9765403Z     QtYARPScopePlugin.vcxproj -> D:\a\robotology-superbuild\robotology-superbuild\build\src\YARP\lib\qt5\qml\Release\robotology\yarp\scope\QtYARPScopePlugin.dll
2021-02-25T11:24:16.7962786Z ##[warning]You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 91 MB
2021-02-25T11:24:17.2776661Z     Building Custom Rule D:/a/robotology-superbuild/robotology-superbuild/src/YARP/src/libYARP_sig/src/CMakeLists.txt
2021-02-25T11:24:17.3763189Z     Image.cpp
~~~

We did not catched this before because only on new Releases the CI job builds both Debug and Release versions of the library. We need a quick fix for this, and then more periodic jobs (that generate the installer without uploading it) to detect this problems happening before of the actual release.

cc @Nicogene

traversaro:
I tried a quick fix by moving the build directory in C: that should have more space, let's see if it works: https://github.com/robotology/robotology-superbuild/releases/tag/v2021.02 .

traversaro:
> 
> 
> I tried a quick fix by moving the build directory in C: that should have more space, let's see if it works: https://github.com/robotology/robotology-superbuild/releases/tag/v2021.02 .

The fix was successful and the installer were generated, now we just need to propagate it to the master branch.

PR #640, opened 2021-02-27
GitHub Actions CI: On Windows build on C:\

traversaro:
The C:\ drive has much more space then the D:\ drive that is used by default.

Fix https://github.com/robotology/robotology-superbuild/issues/637 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,6 +194,25 @@ jobs:
             project_tags_cmake_options: "-DROBOTOLOGY_PROJECT_TAGS=Custom -DROBOTOLOGY_PROJECT_TAGS_CUSTOM_FILE=${GITHUB_WORKSPACE}/releases/2021.02.yaml"
     steps:
     - uses: actions/checkout@master
+    
+    - name: Move robotology-superbuild in C under Windows
+      if: contains(matrix.os, 'windows')
+      shell: bash
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -194,6 +194,25 @@ jobs:
             project_tags_cmake_options: "-DROBOTOLOGY_PROJECT_TAGS=Custom -DROBOTOLOGY_PROJECT_TAGS_CUSTOM_FILE=${GITHUB_WORKSPACE}/releases/2021.02.yaml"
     steps:
     - uses: actions/checkout@master
+    
+    - name: Move robotology-superbuild in C under Windows
+      if: contains(matrix.os, 'windows')
+      shell: bash
+      run: |
+        ls /d/
+        cp -r ${GITHUB_WORKSPACE} /c/
```
Nicogene, line 203: What about `mv` instead of `cp` ?

Nicogene (approved):
A minor comment, LGTM

.github/workflows/ci.yml
traversaro: I tried, and it was complaining that the directory was used (probably there is some process monitoring its content), so that's the reason why I went for copying it.

Nicogene: Ok nevermind 👍🏻

PR #642, opened 2021-02-28
Buildbipedal-locomotion-framework.cmake: Set FRAMEWORK_USE_YARP and Buildbipedal-locomotion-framework.cmake option to ON

traversaro:
If for any reason or bug YARP or matio-cpp or one of its dependencies are missing, then bipedal-locomotion-framework built by the robotology-superbuild should fail with a clear error, not silently disable the relative support. 

See https://github.com/dic-iit/matio-cpp/pull/41#issuecomment-787496467 for a background on this fix.


traversaro:
macOS/Homebrew failures are unrelated, see https://github.com/traversaro/github-actions-brew-update-upgrade-check/actions/runs/608289319 . macOS/conda by the way works fine instead.

S-Dafarra (approved):

Issue #620, opened 2021-02-09
Generate conda binaries and publish them in a robotology conda channel


traversaro:
As https://github.com/robotology/robotology-superbuild/issues/477 is now fixed and the documentation on how to build the robotology-superbuild from source with conda-forge dependencies is available for the interested user at https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-forge.md, I think we can now track in this issue the work to automatically provide conda **binaries** for the packages in the robotology-superbuild, especially because some PIs extressed interest in this activity @lornat75 @DanielePucci .

## Goals

The main objective of this are: 
* Speedup the CI process for projects that depend on a lot of projects that are contained in the robotology-superbuild, without the need to manually maintain CI caches 
* Speedup the time spent in creation of Docker images, also for use in cloud system such as [GitPod](https://www.gitpod.io/) or [GitHub Codespaces](https://github.com/features/codespaces)
* Simplify the installation process of robotology software, especially for MATLAB/Simulink or Python users without a lot of experience of CMake/C++ tools, or users on platforms in which a C++ compiler is not easily installable
* Simplify installation of robotology software in cloud-based systems such as [Binder](https://github.com/binder-examples/conda) or [Google Colab](https://towardsdatascience.com/conda-google-colab-75f7c867a522)
* Simplify generation of Windows installers using [conda Constructor](https://github.com/conda/constructor), to avoid the error-prone current process of compiling every time from source all the dependencies with vcpkg (the CI https://github.com/robotology/robotology-superbuild-dependencies-vcpkg at the moment takes 9/10 hours, so debugging any problem is quite complicated)
* Simplify quick installation of multiple version of the libraries side-by-side for debugging problems (my actual use case : ) and the reason I work on this )

## Strategy

The basic idea is to setup a [conda channel](https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/channels.html) that builds on top of [`conda-forge`](https://conda-forge.org/), and contains the packages installed by the robotology-superbuild, without requiring for package maintainers any additional step beside adding their project to the robotology-superbuild. This idea of a "conda channel" that builds on top of conda-forge has been already successfully exploited in several scientific domains:
* [Bioconda](https://bioconda.github.io/) for Biological sciences, see their [Nature paper](https://www.nature.com/articles/s41592-018-0046-7)
* [RoboStack](https://github.com/RoboStack/ros-noetic) that packages ROS software for conda

## Prototype

[...]

PR #651, opened 2021-03-08
 Add detailed recommended instructions on how to install miniforge3 conda distribution

traversaro:
While the instructions for building from source (and soon installing from binaries, see https://github.com/robotology/robotology-superbuild/issues/620) the robotology-superbuild packages using conda-forge-provided dependencies should work fine on all possible conda distributions, for all users that are new to conda we suggest to use the [miniforge distribution](https://github.com/conda-forge/miniforge). To avoid confusions, we add some detailed documentation on how to install this conda distribution.

### DIFF
--- a/doc/conda-forge.md
+++ b/doc/conda-forge.md
@@ -26,7 +26,7 @@ This section describes how to compile and install the robotology-superbuild with
 
 ### Install a conda distribution
 If you do not have a conda distribution on your system, we suggest to use the minimal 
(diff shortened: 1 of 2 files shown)


traversaro:
@GiulioRomualdi as you raised this point, it would be great if you could look into the docs and check if it make sense and it is understandable by someone without conda experience.

GiulioRomualdi (approved):
That's super! thank you @traversaro

PR #652, opened 2021-03-09
Add automatic generation of conda binary packages for Windows, macOS and Linux in robotology channel

traversaro:
This PR fixes https://github.com/robotology/robotology-superbuild/issues/620 . In particular, it adds machinery to generate the **conda recipes for each subproject of the superbuild**, and a GitHub Action job that build binaries once a week for all the generated recipes for the latest release of the robotology-superbuild projects and loads them in the [**robotology conda channel**](https://anaconda.org/robotology), that builds on top of the [`conda-forge`](https://conda-forge.org/) similarly to who the [`robostack`](https://github.com/RoboStack/ros-noetic) channel for ROS software build on top of [`conda-forge`](https://conda-forge.org/) as well.

It is unfortunately a relatively large PR, but I preferred to experiment with it a bit to understand there is something functional before adding it to the official robotology-superbuild repo. In particular, the test binaries were generated in the channel https://anaconda.org/robotology-staging . The modifications can be largely divided in 5 main categories:
1. Changes in the `conda` directory, that contain all the machinery required to generate conda recipes out for all superbuild projects. These internal mechanism are described in [`doc/conda-recipe-generation.md`](https://github.com/robotology/robotology-superbuild/pull/652/files#diff-1448456855abb123feefa353fc9f53bd6da5f2a085c55da7107bfa3fa554e947) document
2. The GitHub Action generate-conda-packages ( `.github/workflows/generate-conda-packages.yaml` ) that uses the machinery described before to actually compile the conda binaries and upload them to [**robotology conda channel**](https://anaconda.org/robotology)
3. Modifications in the `cmake/Build***.cmake`  scripts, that are basically just adding metadata required to generate conda packages 
4. Modification in developer-oriented documentation in ` doc/developers-faqs.md` (https://github.com/robotology/robotology-superbuild/pull/652/files#diff-f9b520063217385eaf8b07ab7805449c26b59ccb1158241d51497bff72e92100) that are oriented towards package and profile mantainers describing how to add the required conda metadata when they add a new package
5. Modification in user-oriented documentation in `doc/conda-forge.md` ( https://github.com/robotology/robotology-superbuild/pull/652/files#diff-58f523b9fc5050a95da9ce381d330a2cf3ad7f1c88e6fb407fad9fd87d09d091 ): this are meant for end-users of robotology software that want to install the robotology packages directly in binary format. As this is still experimental and the `robotology` channel is still empty, this documentation is still **not linked** in the main README.

[...]

PR #665, opened 2021-03-24
Add option to run generate-conda-packages job without uploading and fix generate-conda-packages

traversaro:
After merging https://github.com/robotology/robotology-superbuild/pull/652, I tried to run `generate-conda-packages` via workflow_dispatch and it was failing due to this:
~~~ 
-- Saved metametadata in /home/runner/work/robotology-superbuild/robotology-superbuild/build/conda/robotology-superbuild-conda-metametadata.yaml
python: can't open file '/home/runner/work/robotology-superbuild/robotology-superbuild/conda/python/generate_conda_recipes_from_metadatadata.py': [Errno 2] No such file or directory
CMake Error at conda/cmake/RobotologySuperbuildGenerateCondaRecipes.cmake:154 (message):
  Error in execution of script
  /home/runner/work/robotology-superbuild/robotology-superbuild/conda/python/generate_conda_recipes_from_metadatadata.py
  conda/cmake/RobotologySuperbuildGenerateCondaRecipes.cmake:195 (generate_conda_recipes)
  CMakeLists.txt:43 (include)
~~~

Probably some change occured doing the review that broke the build. However, this revealed the need to be able to test this github actions on PRs, but without uploading any package. For doing so, I also modified the job to have an input to the `workflow_dispatch`.

The actual conda package generation should be fixed by https://github.com/robotology/robotology-superbuild/commit/82d4cdbe869cebe22327923ff712ba431d96fe01, but let's try it first.

Example of invocation of a job:
workflow_dispatch


traversaro:
Please ignore the failure in the normal C++ job, I killed it as it is not relevant for this PR, the releant job that I  triggered manually is: https://github.com/robotology/robotology-superbuild/runs/2182648284 .

traversaro:
> The actual conda package generation should be fixed by [82d4cdb](https://github.com/robotology/robotology-superbuild/commit/82d4cdbe869cebe22327923ff712ba431d96fe01), but let's try it first.



> 
> 
> Please ignore the failure in the normal C++ job, I killed it as it is not relevant for this PR, the relevant job that I triggered manually is: https://github.com/robotology/robotology-superbuild/runs/2182648284 .

The build completed successfully for Linux ( https://github.com/robotology/robotology-superbuild/runs/2182648284 ) and is ongoing for Windows and macOS, so I think it is safe to ask for a review now.

conda/cmake/RobotologySuperbuildGenerateCondaRecipes.cmake
```diff
@@ -142,7 +142,7 @@ macro(generate_metametadata_file)
 endmacro()
 
 macro(generate_conda_recipes)
-  set(python_generation_script "${CMAKE_CURRENT_SOURCE_DIR}/conda/python/generate_conda_recipes_from_metadatadata.py")
+  set(python_generation_script "${CMAKE_CURRENT_SOURCE_DIR}/conda/python/generate_conda_recipes_from_metametadata.py")
```
prashanthr05, line 145: Just a curiosity, why is it called `metametadata`?  Curious about the two metas ;D

prashanthr05 (approved):

conda/cmake/RobotologySuperbuildGenerateCondaRecipes.cmake
traversaro: A conda recipe contains a file `meta.yaml` that is supposed to contain metadata of the package being built (see https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html). In our case, this meta.yaml file is generated from a template (https://github.com/robotology/robotology-superbuild/blob/master/conda/recipe_template/meta.yaml) using [Jinja2](https://jinja.palletsprojects.com/en/2.11.x/) template engine.
This template so is used to generate the `meta.yaml` file that contains **meta**data, but it does so by getting the relevant info that so contains **metametadata**, because it is metadata about the metadata. This **memetametadata** file itself is generated automatically by the CMake logic of the superbuild, for more on this feel free to refer to https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-recipe-generation.md#internals . Actually I noticed a typo also there.

PR #676, opened 2021-03-30
Fix upload of Conda packages triggered by schedule

traversaro:
There was a bug in the condition for which conda packages are uploaded, in particular for scheduled jobs the generated packages were not uploaded (see https://github.com/robotology/robotology-superbuild/actions/runs/699934415 for an example).


traversaro:
Friendly ping @nunoguedelha .

nunoguedelha (approved):
LGTM. Sorry for the delay @traversaro .

Side note: I have read [this github documentation](https://docs.github.com/en/actions/reference/events-that-trigger-workflows) to get acquainted with the topic.

PR #677, opened 2021-04-06
Add YARP_telemetry to the Dynamics profile

Nicogene:
As per the title.


cmake/BuildYARP_telemetry.cmake
```diff
@@ -0,0 +1,22 @@
...
+              REPOSITORY robotology/yarp-telemetry.git
+              TAG master
+              COMPONENT dynamics
+              FOLDER src
+              DEPENDS YCM
+                      YARP
+                      matioCpp)
+
+set(YARP_telemetry_CONDA_DEPENDENCIES boost-cpp)
+set(YARP_telemetry_CONDA_VERSION "2.3.0.1")
```
traversaro, line 22: This is an override that should not be necessary if the correct version is added in latest releases file, did you added it for a specific reason or it is just a copy&paste leftover?

Nicogene: Do you mean line 22?
Actually, it is a copy-paste and I wanted to ask you what those variables are fore 😄

traversaro: A copy-paste from where?

traversaro: Some docs are available in https://github.com/robotology/robotology-superbuild/blob/master/doc/developers-faqs.md#how-to-ensure-that-binary-packages-are-correctly-generated-for-a-new-package, but probably the docs on version are missing.

Nicogene: >A copy-paste from where?

From `blocktestcore`

traversaro: Good catch, indeed the version number is wrong: https://anaconda.org/robotology/blocktest .

traversaro:
@Nicogene I bumped the used version of robotology/robotology-superbuild-dependencies-vcpkg in https://github.com/robotology/robotology-superbuild/pull/677/commits/c02873ed6b9a49d75748a9aba693b4de30532438 . Can you update the CHANGELOG describing this change? Thanks!

traversaro (approved):

traversaro:
Thanks @Nicogene and @AlexAntn !

Issue #681, opened 2021-04-10
Conda Linux CI build fail due to linking problem in Wearables


traversaro:
The last two nightly jobs failed on Conda/Linux with the following error:
~~~
2021-04-10T02:58:19.0574598Z [290/328] Performing build step for 'wearables'
2021-04-10T02:58:19.0575991Z FAILED: src/wearables/CMakeFiles/YCMStamp/wearables-build 
2021-04-10T02:58:19.0578260Z cd /home/runner/work/robotology-superbuild/robotology-superbuild/build/src/wearables && /usr/share/miniconda/envs/test/bin/cmake --build . && /usr/share/miniconda/envs/test/bin/cmake -E touch /home/runner/work/robotology-superbuild/robotology-superbuild/build/src/wearables/CMakeFiles/YCMStamp/wearables-build
2021-04-10T02:58:19.0580224Z [1/66] Generating code from thrift/XsensSuitControlService.thrift
2021-04-10T02:58:19.0581275Z [2/66] Generating code from thrift/WearableActuators.thrift
2021-04-10T02:58:19.0582369Z [3/66] Building CXX object msgs/CMakeFiles/WearableActuators.dir/thrift/ActuatorStatus.cpp.o
2021-04-10T02:58:19.0583556Z [4/66] Building CXX object msgs/CMakeFiles/WearableActuators.dir/thrift/ActuatorType.cpp.o
2021-04-10T02:58:19.0584858Z [5/66] Building CXX object msgs/CMakeFiles/XsensSuitControl.dir/thrift/XsensSuitControlService.cpp.o
2021-04-10T02:58:19.0586156Z [6/66] Building CXX object msgs/CMakeFiles/WearableActuators.dir/thrift/ActuatorInfo.cpp.o
2021-04-10T02:58:19.0587287Z [7/66] Building CXX object impl/SensorsImpl/CMakeFiles/SensorsImpl.dir/SensorsImpl.cpp.o
2021-04-10T02:58:19.0588252Z [8/66] Linking CXX shared library lib/libSensorsImpl.so
2021-04-10T02:58:19.0589196Z [9/66] Building CXX object msgs/CMakeFiles/WearableActuators.dir/thrift/Haptic.cpp.o
2021-04-10T02:58:19.0590273Z [10/66] Building CXX object msgs/CMakeFiles/WearableActuators.dir/thrift/Motor.cpp.o
[...]

PR #682, opened 2021-04-10
Add "Print Used Environment" step to conda CI

traversaro:
In this list, two commands are executed:
* `mamba list`, to print the exact packages installed in the conda environment
* `env` to print the environment variables set

This are useful to quickly check what changed in case of CI failures (such as https://github.com/robotology/robotology-superbuild/issues/681), as it is simplify the process of checking the difference to just copy&paste the output of that step in meld or any other diff software.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,7 @@ jobs:
       shell: bash -l {0}
       run: |
         mamba list
+        env
 
     - name: Configure [Conda/Linux&macOS]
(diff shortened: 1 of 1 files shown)


Nicogene (approved):

nunoguedelha:
Hi @traversaro , the commands `env` and `conda info -s` both return the list of environment variables. I noticed that the outputs are slightly different. since the outputs of `env` and `echo` in a terminal for a given variable still match, I guess that `conda info -s` proides some kind of preset/default values?

Issue #679, opened 2021-04-06
Installation error: Could not find a package configuration file provided by "YARP_robotinterface"


Jingyue21:
Dear developers,

I'm installing iCub software following the [instructions](https://github.com/robotology/robotology-superbuild#linux-from-source). 
My system is Ubuntu 18.04.5 LTS (Bionic Beaver) with cmake version 3.18.2. I'm installing Distro 2021.02. When I did superbuild->ccmake, I turned on "ROBOTOLOGY_USES_PYTHON", "ROBOTOLOGY_ENABLE_EVENT_DRIVEN" and "ROBOTOLOGY_USES_GAZEBO" apart from the default configurations.
When I did superbuild->make, I got the error below. Could you kindly help me to solve this error? Many thanks in advance!

[ 51%] Completed 'icub-models'
[ 57%] Built target icub-models
[ 58%] Skipping update step for 'GazeboYARPPlugins'
[ 60%] Performing configure step for 'GazeboYARPPlugins'
Not searching for unused variables given on the command line.
loading initial cache file /<mypath>/robotology-superbuild/build/src/GazeboYARPPlugins/CMakeFiles/YCMTmp/GazeboYARPPlugins-cache-Release.cmake
CMake Error at /usr/local/lib/cmake/YARP/YARPConfig.cmake:162 (find_package):
  Could not find a package configuration file provided by
  "YARP_robotinterface" with any of the following names:

    YARP_robotinterfaceConfig.cmake
    yarp_robotinterface-config.cmake

  Add the installation prefix of "YARP_robotinterface" to CMAKE_PREFIX_PATH
  or set "YARP_robotinterface_DIR" to a directory containing one of the above
  files.  If "YARP_robotinterface" provides a separate development package or
  SDK, be sure it has been installed.
Call Stack (most recent call first):
  CMakeLists.txt:48 (find_package)

-- Configuring incomplete, errors occurred!
See also "/<mypath>/robotology-superbuild/build/src/GazeboYARPPlugins/CMakeFiles/CMakeOutput.log".
CMakeFiles/GazeboYARPPlugins.dir/build.make:120: recipe for target 'src/GazeboYARPPlugins/CMakeFiles/YCMStamp/GazeboYARPPlugins-configure' failed
make[2]: *** [src/GazeboYARPPlugins/CMakeFiles/YCMStamp/GazeboYARPPlugins-configure] Error 1
[...]

PR #683, opened 2021-04-11
In conda docs explicitly indicate when the use of conda-forge channel  is needed

traversaro:
When building from sources or from binaries, we were implicitly assuming that the user already had the `conda-forge` channel configured by default. While this is the case for the `miniforge` distribution that we **suggest** to use, it may not be true for users or system using other distributions such as miniconda (see for example https://github.com/dic-iit/bipedal-locomotion-framework/issues/230#issuecomment-815562835) or anaconda (https://github.com/robotology/robotology-superbuild/issues/679).

For this reason, it make sense to modify the instructions to always specify the `-c conda-forge` option. This is also useful for the future, when we may migrate packages from the `robotology` channel to the `conda-forge` one, and we want the `conda-forge` one to have the precedence.

### DIFF
--- a/doc/conda-forge.md
+++ b/doc/conda-forge.md
@@ -55,24 +55,24 @@ conda activate robotologyenv
 
 Once you are in an activated environment, you can install robotology packages by just running the command:
 ~~~
-conda install -c robotology <packagename>
+conda install -c conda-forge -c robotology <packagename>
 ~~~
 
 The list of available packages is available at https://anaconda.org/robotology/repo . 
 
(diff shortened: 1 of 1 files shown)


traversaro:
cc @isorrentino @GiulioRomualdi @Jingyue21

traversaro:
Please ignore the conda-forge CI failure, it should be fixed by https://github.com/robotology/wearables/pull/110 .

GiulioRomualdi (approved):

PR #684, opened 2021-04-11
Do not source hardcoded Gazebo's setup.sh on Conda

traversaro:
On conda, the Gazebo's environment variables are already set in the environment activation scripts, so there is no need to source it manually in the `setup.sh` script. Actually, we need to make sure that the apt-installed Gazebo setup.sh is not accidentally sourced, as that could conflict with the conda-installed one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,10 +61,10 @@ endif()
 
 # Configure a setup.sh
 if(ROBOTOLOGY_USES_GAZEBO)
-  if(WIN32)
-    # On Windows we assume that the Gazebo's enviroment variables are already set,
+  if(WIN32 OR ROBOTOLOGY_CONFIGURING_UNDER_CONDA)
+    # On Windows or Conda we assume that the Gazebo's enviroment variables are already set,
     # so we set as this name a file that does not exists and so will be ignored
-    set(GAZEBO_SETUP_SH_PATH "on-windows-we-assume-that-gazebo-env-variables-are-already-set")
+    set(GAZEBO_SETUP_SH_PATH "on-windows-or-conda-we-assume-that-gazebo-env-variables-are-already-set")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -61,10 +61,10 @@ endif()
 
 # Configure a setup.sh
 if(ROBOTOLOGY_USES_GAZEBO)
-  if(WIN32)
-    # On Windows we assume that the Gazebo's enviroment variables are already set,
+  if(WIN32 OR ROBOTOLOGY_CONFIGURING_UNDER_CONDA)
+    # On Windows or Conda we assume that the Gazebo's enviroment variables are already set,
     # so we set as this name a file that does not exists and so will be ignored
-    set(GAZEBO_SETUP_SH_PATH "on-windows-we-assume-that-gazebo-env-variables-are-already-set")
+    set(GAZEBO_SETUP_SH_PATH "on-windows-or-conda-we-assume-that-gazebo-env-variables-are-already-set")
```
nunoguedelha, line 67: Side note: I guess that you've chosen to set a message string here instead of leaving the variable empty because the user gets a "warning" if he tries to use the variable?

traversaro: Yes, that was the idea.

nunoguedelha (commented):
Just left a minor comment regarding the dummy value of GAZEBO_SETUP_SH_PATH.

traversaro:
> 
> 
> Just left a minor comment regarding the dummy value of GAZEBO_SETUP_SH_PATH.

Ack, do you think something can be improved w.r.t. to that?

nunoguedelha:
> > Just left a minor comment regarding the dummy value of GAZEBO_SETUP_SH_PATH.
> 
> Ack, do you think something can be improved w.r.t. to that?

No no it's fine to me.

nunoguedelha (approved):

Issue #695, opened 2021-04-22
All builds failing on Unstable branches


traversaro:
Related issue: https://github.com/robotology/yarp/issues/2556 .

Issue #693, opened 2021-04-22
Windows/Conda and macOS/Homebrew build failed on Stable and Unstable branches


traversaro:
It seems that the changes in https://github.com/robotology/icub-basic-demos/pull/32 broke the compilation on:
* Windows/Conda 
* macOS/Homebrew 

Probably this is due to some Gazebo/problem bug, as instead the compilation (with Gazebo enabled) of Windows/vcpkg and macOS/Conda is working fine. Also, the CMake in gazebo-yarp-plugins seems to be working fine instead on those platforms, see https://github.com/robotology/gazebo-yarp-plugins/blob/40ef333672b9cc85cdf50255d229b6bb462237cb/CMakeLists.txt#L49 and https://github.com/robotology/gazebo-yarp-plugins/blob/40ef333672b9cc85cdf50255d229b6bb462237cb/plugins/controlboard/CMakeLists.txt#L42 .

traversaro:
Example failure (see https://github.com/robotology/robotology-superbuild/actions/runs/772709209):

## Windows/Conda
**fatal error LNK1181: cannot open input file 'OgreBites.lib** 
~~~
2021-04-22T03:45:33.6354634Z C:\Miniconda\envs\test\Library\include\gazebo-11\gazebo/physics/Link.hh(691,52): warning C4251: 'gazebo::physics::Link::dataPtr': class 'std::unique_ptr<gazebo::physics::LinkPrivate,std::default_delete<gazebo::physics::LinkPrivate>>' needs to have dll-interface to be used by clients of class 'gazebo::physics::Link' [D:\a\robotology-superbuild\robotology-superbuild\build\src\icub-basic-demos\demoRedBall\demoRedBall-world.vcxproj] [D:\a\robotology-superbuild\robotology-superbuild\build\icub-basic-demos.vcxproj]
2021-04-22T03:45:33.6358949Z   C:\Miniconda\envs\test\Library\include\gazebo-11\gazebo/physics/Link.hh(691): message : see declaration of 'std::unique_ptr<gazebo::physics::LinkPrivate,std::default_delete<gazebo::physics::LinkPrivate>>' [D:\a\robotology-superbuild\robotology-superbuild\build\src\icub-basic-demos\demoRedBall\demoRedBall-world.vcxproj]
[...]

PR #697, opened 2021-04-23
Specify both YARP 3.4 and 3.5 style options for depthimage portmonitor

traversaro:
See https://github.com/robotology/yarp/pull/2552 . As we now need to support both YARP 3.4 (Stable branches and latest releases) and YARP 3.5 (Unstable branches) we will pass both options, as this should work fine and avoid to trigger the bug https://github.com/robotology/yarp/issues/2556 .

Fix https://github.com/robotology/robotology-superbuild/issues/695


diegoferigo (approved):

traversaro:
Some CI jobs are still failing, but this is due to https://github.com/robotology/robotology-superbuild/issues/693 that hopefully should be fixed by https://github.com/robotology/icub-basic-demos .

Nicogene (approved):

Issue #698, opened 2021-04-24
Regression in macOS/Conda builds


traversaro:
It seems that https://github.com/robotology/robotology-superbuild/pull/694  introduced a regression, that went unnoticed due to https://github.com/robotology/robotology-superbuild/issues/695 and https://github.com/robotology/robotology-superbuild/issues/693  . 

The CasADi Python bindings are failing with error:
~~~
2021-04-24T02:46:53.2116930Z /usr/local/miniconda/envs/test/bin/x86_64-apple-darwin13.4.0-clang++ -DCASADI_VERSION=31 -DHAVE_MKSTEMPS -DUSE_CXX11 -DWITH_DEEPBIND -DWITH_DEPRECATED_FEATURES -DWITH_DL -D_SCL_SECURE_NO_WARNINGS -D_USE_MATH_DEFINES -D_casadi_EXPORTS -I/Users/runner/work/robotology-superbuild/robotology-superbuild/src/casadi/. -I. -I/Users/runner/work/robotology-superbuild/robotology-superbuild/src/casadi/swig/.. -I/Users/runner/work/robotology-superbuild/robotology-superbuild/src/casadi/swig/python -I/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7 -Icasadi/core/runtime -I/Users/runner/work/robotology-superbuild/robotology-superbuild/src/casadi/casadi/core/../.. -Icasadi/core/../.. -march=core2 -mtune=haswell -mssse3 -ftree-vectorize -fPIC -fPIE -fstack-protector-strong -O2 -pipe -stdlib=libc++ -fvisibility-inlines-hidden -std=c++14 -fmessage-length=0 -isystem /usr/local/miniconda/envs/test/include -fPIC -O3 -DNDEBUG -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk -fPIC -Wno-dynamic-class-memaccess -Wno-self-assign -MD -MT swig/python/CMakeFiles/_casadi.dir/__/casadiPYTHON_wrap.cxx.o -MF swig/python/CMakeFiles/_casadi.dir/__/casadiPYTHON_wrap.cxx.o.d -o swig/python/CMakeFiles/_casadi.dir/__/casadiPYTHON_wrap.cxx.o -c swig/casadiPYTHON_wrap.cxx
2021-04-24T02:46:53.2211340Z In file included from swig/casadiPYTHON_wrap.cxx:191:
[...]

PR #694, opened 2021-04-22
Add the possibility to compile the casadi python bindings

GiulioRomualdi:
Before merging the PR we should use casadi version `3.5.5.3`. I've just created the tag. I don't know if I can bump the new tag in this PR or handle it in a separate PR

PR #699, opened 2021-04-24
Buildcasadi: pass WITH_PYTHON3 option when Python bindings are built

traversaro:
Follow up of https://github.com/robotology/robotology-superbuild/pull/694, fix https://github.com/robotology/robotology-superbuild/issues/698 by explicitly requesting to use Python3, that means that the system Python should not be used on macOS .

The use of this option is copied from https://github.com/conda-forge/casadi-feedstock/blob/391139c859e1fba1321818bae27b6ca11fa4081e/recipe/build.sh#L7 .


GiulioRomualdi (approved):

traversaro:
I also added a necessary option `WITH_COPYSIGN_UNDEF` that is needed when compiling the Python bindings on Windows.

PR #701, opened 2021-04-25
Add bump-conda-build-number job to generate-conda-packages

traversaro:
To ensure that each new build of the packages get correctly uploaded, even if the package version was not uploaded (for example because one of the dependencies that influences its ABI was updated) we have a `CONDA_BUILD_NUMBER` parameter in the `conda/cmake/CondaGenerationOptions.cmake` , that specifies the [Conda build number](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#build-number-and-string) that is used by all Conda packages.

This parameter should be bumped at each successful generation and upload of Conda binaries, and until now I have done this manually (see for example https://github.com/robotology/robotology-superbuild/commit/7a10e5088ad33a2166bb5635038b9c58fbb5fc8c and https://github.com/robotology/robotology-superbuild/commit/0df5e6a997ce716a947ef09630516db5b1726e12). However, this is quite error prone, so in this PR I added an additional job that if the generation and upload of packages works fine, also bumps the `CONDA_BUILD_NUMBER`  automatically.

### DIFF
--- a/.github/workflows/generate-conda-packages.yaml
+++ b/.github/workflows/generate-conda-packages.yaml
@@ -117,4 +117,17 @@ jobs:
             ls *.tar.bz2
             anaconda upload --skip-existing *.tar.bz2
 
+    # If the  generate-conda-packages completed correctly and binaries are uploaded,
+    # bump automatically the CONDA_BUILD_NUMBER in conda/cmake/CondaGenerationOptions.cmake
+    # for future builds
+    bump-conda-build-number:
+        name: "Bump Conda Build number for future builds"
+        runs-on: ubuntu-latest
+        needs: generate-conda-packages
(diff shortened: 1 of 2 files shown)


traversaro:
The macOS failure is unrelated and already fixed by https://github.com/robotology/robotology-superbuild/pull/699 .

diegoferigo (approved):

traversaro:
This failed because:
> Error: Error: Pushing to https://github.com/robotology/robotology-superbuild
POST git-receive-pack (717 bytes)
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: At least 1 approving review is required by reviewers with write access.        
error: failed to push some refs to 'https://github.com/robotology/robotology-superbuild'

However, for this kind of commits there is no reason to require a review.

PR #730, opened 2021-05-13
Add YARP_telemetry dependency for wearables

prashanthr05:
The PR in `wearables` https://github.com/robotology/wearables/pull/113 adds `YARP_telemetry` as one of its dependency. 

This PR complements that change.

### DIFF
--- a/cmake/Buildwearables.cmake
+++ b/cmake/Buildwearables.cmake
@@ -6,6 +6,7 @@ include(FindOrBuildPackage)
 
 find_or_build_package(YARP QUIET)
 find_or_build_package(iDynTree QUIET)
+find_or_build_package(YARP_telemetry QUIET)
 
 set(WEARABLES_CMAKE_ARGS "")
 if(WIN32)
@@ -20,4 +21,5 @@ ycm_ep_helper(wearables TYPE GIT
               FOLDER src
               DEPENDS YARP
                       iDynTree
+                      YARP_telemetry
               CMAKE_ARGS ${WEARABLES_CMAKE_ARGS})


prashanthr05:
cc @traversaro

prashanthr05:
I accidentally closed the PR. Reopened!

cmake/Buildwearables.cmake
```diff
@@ -6,6 +6,7 @@ include(FindOrBuildPackage)
 
 find_or_build_package(YARP QUIET)
 find_or_build_package(iDynTree QUIET)
+find_or_build_package(YARP_telemetry QUIET)
```
traversaro, line 9: You need to also add `YARP_telemetry` to the `DEPENDS` argument of `ycm_ep_helper` (probably we should eventually  modify `ycm_ep_helper` to avoid having two places where to list the dependency.

prashanthr05: I tried compiling without that in my local machine, looks like the `wearables` is compiling fine with IWearLogger enabled.

prashanthr05: Maybe because it contained as one of the components in YARP?

traversaro: The DEPENDS just ensure that `wearables` is build **after** `YARP_telemetry`. It is perfectly possible that in your case, by chance  `YARP_telemetry` is built and installed before `wearables` and so everything works fine, but we should add `YARP_telemetry` to ensure that this happens.

prashanthr05: Ah understood.

prashanthr05: I rebased with the suggested change in https://github.com/robotology/robotology-superbuild/pull/730/commits/a6931f3e7da85cb1b5fb997979b2bfce59c7af1f

traversaro (approved):
A lot of CI failures, all unrelated.

PR #743, opened 2021-05-19
Add Release Github Action

### LINKED ISSUE #636: Create a manually dispatched GitHub Action to do a new release
Doing a  new release is not a time intensive action (a few minutes every 3 months) but it is quite error prone, as we keep forgetting steps (see https://github.com/robotology/robotology-superbuild/issues/635) and there are some steps that are intrinsically prone to errors (for example, when tagging a new release on the releases/yyyy.mm branch, the default setting of the GitHub UI is to do tag the release on `master` branch, and it is quite easy to forget to switch to releases/yyyy.mm branch). 

For this reason, it may be convenient to create a GitHub Action to do the steps documented in https://github.com/robotology/robotology-superbuild/blob/v2021.02/doc/developers-faqs.md#how-to-do-a-new-release automatically.

Nicogene:
This PR add a github action for creating a new `robotology-superbuild` release, it automatize these steps: https://github.com/robotology/robotology-superbuild/blob/master/doc/developers-faqs.md#how-to-do-a-new-release

For doing this I changed the CMakeLists of windows packaging https://github.com/robotology/robotology-superbuild/commit/a2e8999f97d87940dd0c527266e528d711a1b215 for make easier the sed command, the idea is that on master the `INSTALLER_VERSION` is empty(or we can put something like  `master` and then it is versioned in the `releases/yyyy.mm` branch.

This Action can be triggered only by hand specifying the version of the release you want to create:

immagine

Here are the steps:

- Copy `latest.yml` into `yyyy.mm.yml` commit and push it on master https://github.com/Nicogene/robotology-superbuild/commit/0efd5cb406d7d9e4eafbbba4b6f314f8d22ff6ab
- Switch then on [`releases/yyyy.mm`](https://github.com/Nicogene/robotology-superbuild/tree/releases/2021.05) and does these 2 commits https://github.com/Nicogene/robotology-superbuild/commit/96279a0de5c7249e55f9935f54d0297fbb8d1e49 and https://github.com/Nicogene/robotology-superbuild/commit/e3b9b7d6953b9a4c8ef767f142e1a4df7383ab20
- Tag on that branch https://github.com/Nicogene/robotology-superbuild/releases/tag/v2021.05

It fixes #636

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,51 @@
+name: Create a new release
+
+on:
+  workflow_dispatch:
+    inputs:
+      superbuild_tag:
+        description: 'The new robotology-superbuild tag(without the "v")'
+        required: true
+jobs:
+  PrepareRelease:
+    name: Create the ${{ github.event.inputs.superbuild_tag }} release
+    runs-on: [ubuntu-latest]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Modify the yaml file
+      run: |
+        cd releases
+        cp ./latest.releases.yaml ./${{ github.event.inputs.superbuild_tag }}.yaml
+
+    - name: Configure Git
+      run: |
+        git config --global push.default upstream
+        git config --global user.name "GitHub Actions"
+        git config --global user.email "actions@github.com"
+        git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}.git
+
+    - name: Commit and push the ${{ github.event.inputs.superbuild_tag }} yaml
+      run: |
+        git add .
+        git commit -m "Copying latest.releases.yaml to ${{ github.event.inputs.superbuild_tag }}.yaml"
+        git push
+
+    - name: Prepare and push ${{ github.event.inputs.superbuild_tag }} branch
+      run: |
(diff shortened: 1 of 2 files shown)


traversaro:
Great. Do you want to update https://github.com/robotology/robotology-superbuild/blob/master/doc/developers-faqs.md#how-to-do-a-new-release to mention at the beginning that the documented step can be done via this new github action?

Nicogene:
(The fails are due to a wrong commit in `yarp-telemetry` that has been fixed now)

traversaro:
> Yes, but maybe it is better to add it after merging the PR in this way I can link it in the documentation.

Ok!

traversaro:
> I have only concern about the security of using the `GITHUB_TOKEN`, it inherits the permissions of the guy started the GA? Because since this GA commit push create branches and tag it is not good that everyone can use it (maybe @pattacini has more insight on it)

I guess that you can only trigger the workflow if you already have write access to the repo, but we can check.

pattacini:
>I guess that you can only trigger the workflow if you already have write access to the repo, but we can check.

That's correct for workflow dispatch events.
However, I would also recommend that you enforce security in this case by using [environments](https://docs.github.com/en/actions/reference/environments).

Issue #757, opened 2021-05-24
bipedal-locomotion-framework Python bindings not found if installed via conda binaries on Linux or macOS


traversaro:
If I install blf via conda binaries with Python support (see https://github.com/robotology/robotology-superbuild/pull/749):
~~~
conda create -n robpy
conda activate robpy
conda install -c conda-forge -c robotology python=3.<minor> bipedal-locomotion-framework 
~~~
where `<minor>` can be 6, 7, 8 or 9

Then if I try to run the Python bindings 
~~~
(robpy) traversaro@IITICUBLAP102:~$ python
Python 3.7.10 | packaged by conda-forge | (default, Feb 19 2021, 16:07:37)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import bipedal_locomotion_framework
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'bipedal_locomotion_framework'
>>>
~~~

This is because bipedal-locomotion-framework v0.1.0 install its Python bindings in `$CONDA_PREFIX/lib/python3.<minor>/dist-packages` instead of `export PYTHONPATH=$CONDA_PREFIX/lib/python3.<minor>/site-packages`. Indeed, a temporary workaround is to just define `PYTHONPATH` as:
~~~
export PYTHONPATH=$CONDA_PREFIX/lib/python3.<minor>/dist-packages
~~~~

traversaro:
Let's keep open until we generate good packages.

traversaro:
Windows package generation failed for a transient Network error: https://github.com/robotology/robotology-superbuild/actions/runs/873096279 . However, the issue was just on Linux and macOS, so I think we can close the issue.

PR #758, opened 2021-05-24
blf: Always set FRAMEWORK_DETECT_ACTIVE_PYTHON_SITEPACKAGES to on when generating conda packages

traversaro:
This fixes https://github.com/robotology/robotology-superbuild/issues/757 . The issue would be fixed also as soon as the new version of blf is released, but in any case the logic of installing in the active sitepackage when building a conda package is for sure more robust also against future regressions.

### DIFF
--- a/cmake/Buildbipedal-locomotion-framework.cmake
+++ b/cmake/Buildbipedal-locomotion-framework.cmake
@@ -31,12 +31,12 @@ if (ROBOTOLOGY_ENABLE_DYNAMICS_FULL_DEPS)
 endif()
 
 # For what regards Python installation, the options changes depending
-# on whater we are installing YARP from source, or we are generating a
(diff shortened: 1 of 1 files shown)


diegoferigo (approved):
So, in short, we are using CMake's FindPython3 hoping that it works reliably? With modern versions, luckily conda seems supported and it should work better than a Python system installation since there are a lot of environment variables that point to the right direction.

traversaro:
> So, in short, we are using CMake's FindPython3 hoping that it works reliably?

Actually we completely bypass the find logic by passing the exact Python we want to use in https://github.com/robotology/robotology-superbuild/blob/master/conda/cmake/RobotologySuperbuildGenerateCondaRecipes.cmake#L118 .

Issue #622, opened 2021-02-12
Investigate automatic support for autocomplete integration


traversaro:
YARP (and possibly other projects) support autocomplete integration as discussed in https://github.com/robotology/community/discussions/15 . However, the proposed procedure require to create a symbol link in global directories. To permit automatic and non-invasive integration, we should understand if we enable it on a process-specific level by doing something in our "activation script" setup.sh .

drdanz:
From the readme in https://github.com/scop/bash-completion/

> **Q. Where should I install my own local completions?**
> 
> A. Put them in the `completions` subdir of `$BASH_COMPLETION_USER_DIR`
>   (defaults to `$XDG_DATA_HOME/bash-completion` or
>    `~/.local/share/bash-completion`
>   if `$XDG_DATA_HOME` is not set) to have them loaded on demand.
>   See also the next question's answer for considerations for these
>   files' names, they apply here as well. Alternatively, you can write
>   them directly in `~/.bash_completion` which is loaded eagerly by
>   our main script.

I've never tried if this works, but I think we can put there a file that loads all the files installed in the `<superbuild_install_path>/share/bash-completion/completions/`
Or we could change the `BASH_COMPLETION_USER_DIR` environment variable, but it's more invasive in case the user wants to use his own scripts in that path.

traversaro:
I just used for the first time for a real task the YARP conda binaries, and while they are working fine, the absence of bash completion is a real usability problem.

traversaro:
> 
> 
> From the readme in https://github.com/scop/bash-completion/
> 
> > **Q. Where should I install my own local completions?**
> > A. Put them in the `completions` subdir of `$BASH_COMPLETION_USER_DIR`
> > (defaults to `$XDG_DATA_HOME/bash-completion` or
> > `~/.local/share/bash-completion`
> > if `$XDG_DATA_HOME` is not set) to have them loaded on demand.
> > See also the next question's answer for considerations for these
> > files' names, they apply here as well. Alternatively, you can write
> > them directly in `~/.bash_completion` which is loaded eagerly by
> > our main script.
> 
> I've never tried if this works, but I think we can put there a file that loads all the files installed in the `<superbuild_install_path>/share/bash-completion/completions/`
> Or we could change the `BASH_COMPLETION_USER_DIR` environment variable, but it's more invasive in case the user wants to use his own scripts in that path.

Actually it seems we don't  need to interfere with user-specified completion scripts, the following Q is related to installation:
[...]

PR #759, opened 2021-05-25
Add <superbuild_install_prefix>/share to XDG_DATA_DIRS environment variable to support yarp bash autocompletion out of the box

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/622 .


S-Dafarra (approved):
Cool! Does it work also on Windows?

traversaro:
> 
> 
> Cool! Does it work also on Windows?

I did not tested it, but it could work assuming that you installed `bash-completion` and you use Git Bash. However, in any  case I prefer to set `XDG_DATA_DIRS`  coherently across all platforms as this env variable could also affect other software.

traversaro:
> 
> 
> > Cool! Does it work also on Windows?
> 
> I did not tested it, but it could work assuming that you installed `bash-completion` and you use Git Bash. However, in any case I prefer to set `XDG_DATA_DIRS` coherently across all platforms as this env variable could also affect other software.

If Instead you were referring to to the Powershell-based autocomplete (https://github.com/robotology/yarp/tree/master/data/WindowsPowerShell) I still need to check how to integrate it.

traversaro:
I tested on a real Ubuntu 20.04 box and it seems to work fine!

PR #765, opened 2021-05-28
Update README.md

julijenv:
correct some missing spelling words for windows support


traversaro:
Thanks @julijenv ! If you are interested on installing the robotology software on Windows, you can also check the currently in preview instructions on how to install the software on Windows via conda, in my experience it is much easier (even if not as tested as the classical instructions): https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-forge.md .

traversaro (approved):

Issue #763, opened 2021-05-27
Add nlohmann json to dependencies


traversaro:
It is already used in yarp-telemetry (even if temporary an internal version is used) and will be used in https://github.com/dic-iit/bipedal-locomotion-framework/pull/316/files, so better to add it even becuase it is small and useful library.

traversaro:
The tricky aspect is that `nlohmann-json3-dev` is not available on Ubuntu 18.04, so we can't add it to the long list to copy&paste as we usually do.

PR #776, opened 2021-06-01
Ad nlohmann-json as dependency of the superbuild

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/763 . I added it as a dependency in all points, I am just missing https://github.com/robotology/robotology-superbuild-dependencies-vcpkg/pull/45 to also bump the version of vcpkg dependencies used in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,6 +203,8 @@ jobs:
         apt-get -y update
         apt-get -y upgrade
 
+
+
     - name: Configure [Docker]
       run: |
         mkdir -p build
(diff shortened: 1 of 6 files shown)


traversaro:
> I am just missing [robotology/robotology-superbuild-dependencies-vcpkg#45](https://github.com/robotology/robotology-superbuild-dependencies-vcpkg/pull/45) to also bump the version of vcpkg dependencies used in CI.

This was fixed by https://github.com/robotology/robotology-superbuild/pull/776/commits/de82b26cf2f3c478f157a1f2348d36006b961a95 . As the vcpkg tarball has shown to be more stable lately, I switched CI to also use latest as originally suggested by @S-Dafarra, for consistency with the user docs.

.github/workflows/ci.yml
```diff
@@ -406,6 +413,11 @@ jobs:
         cd build
         cmake -DROBOTOLOGY_ENABLE_DYNAMICS_FULL_DEPS:BOOL=OFF .
 
+    - name: Install dependencies unsupported in Ubuntu 18.04
+      if: contains(matrix.os, '20.04')
```
S-Dafarra, line 417: Why the condition is different here?

traversaro: Because it is for installing dependencies that are not available on 18.04, so it is run on Ubuntu 20.04 .

S-Dafarra: Sorry, I should have been more precise. I meant with respect to the similar if clause above at  line 231

S-Dafarra: I saw later that one applies to docker images and this to os. Not important at all

S-Dafarra (approved):

.github/workflows/ci.yml
traversaro: > 
> 
> I saw later that one applies to docker images and this to os. Not important at all

Yes, that's it. The clause are different as for the docker images you need to check the used docker image, not the base os.

traversaro:
The Conda/Linux CI failure are unrelated (see https://github.com/robotology/robotology-superbuild/issues/785), I think once the other jobs finish we can merge.

Issue #785, opened 2021-06-05
Conda\Linux CI failure due to libsolv 0.7.19 used by mamba


traversaro:
Since Today, the Conda\Linux CI is failing with error:
~~~
 Looking for: ['libignition-gazebo4']

Encountered problems while solving:
  - package python-3.6.12-4_73_pypy requires python_abi 3.6 *_pypy36_pp73, but none of the providers can be installed
~~~

Example of failing job: https://github.com/robotology/robotology-superbuild/pull/784/checks?check_run_id=2753897981 .

traversaro:
Installing `libignition-gazebo4` on an empty environment works fine with both conda and mamba (tested in a local install), so probably this is a problem related to some other package installed earlier by the CI.

traversaro:
By inspecting the logs, it seems that for some reason the step that install regular dependencies:
~~~
mamba install ace asio boost eigen gazebo glew glfw gsl ipopt irrlicht libjpeg-turbo libmatio libode libxml2 nlohmann_json opencv pkg-config portaudio qt sdl sdl2 sqlite tinyxml spdlog
~~~
started to install the Python package provided by pypy 3.7 instead of regular python 3.9 .

traversaro:
I just tried with a local mamba install, and:
~~~
mamba install python
~~~
wants to install Python 3.7 as we would expect.

On the other hand, also locally
~~~
mamba install ace asio boost eigen gazebo glew glfw gsl ipopt irrlicht libjpeg-turbo libmatio libode libxml2 nlohmann_json opencv pkg-config portaudio qt sdl sdl2 sqlite tinyxml spdlog
~~~
results in
~~~
  _libgcc_mutex                       0.1  conda_forge          conda-forge/linux-64     Cached
  _openmp_mutex                       4.5  1_gnu                conda-forge/linux-64     Cached
  ace                               7.0.2  h9c3ff4c_0           conda-forge/linux-64       7 MB
  alsa-lib                          1.2.3  h516909a_0           conda-forge/linux-64     560 KB
  ampl-mp                           3.1.0  h616b090_1004        conda-forge/linux-64       1 MB
  asio                             1.14.1  h9c3ff4c_1           conda-forge/linux-64     333 KB
  boost                            1.74.0  py36hb4f4ce5_3       conda-forge/linux-64     364 KB
  boost-cpp                        1.74.0  hc6e9bd1_3           conda-forge/linux-64      16 MB
[...]

PR #786, opened 2021-06-06
Install boost-cpp instead of boost in conda instructions and CI

traversaro:
This is a correct fix in general, as our use of boost do not depend on boost-python, so just depending on boost-cpp is enough. Furthermore, it also act as a workaround for https://github.com/robotology/robotology-superbuild/issues/785 .


diegoferigo (approved):

traversaro:
Unfortunately, for some strange reason this changes lead to protobuf 3.4 being installed on Windows.

traversaro:
I think we can consider this superseded by https://github.com/robotology/robotology-superbuild/pull/788 .

PR #797, opened 2021-06-14
BuildYARP: Enable all fake** devices

traversaro:
All the fake*** devices are quite useful for tutorials and tests and they do not have strange deps, so let's enable them. 

Let's check CI with attention as it is possible that some of this devices have compilation problems in some platforms tested by CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ### Added
 - Add `nlohmann-json` dependency to the superbuild (https://github.com/robotology/robotology-superbuild/pull/776)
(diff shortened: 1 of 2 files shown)


pattacini (approved):

pattacini (approved):

pattacini:
I think this won't be backported to `2021.05`.
However, I wouldn't expect any problem for the time being given that the training on YARP has been already done today.

This comment is just in view of merging https://github.com/vvv-school/tutorial_yarp-basics/pull/2.

traversaro:
> I think this won't be backported to `2021.05`.

We can backport to 2021.05 if you like, but we would need a new tag in any case to properly propagate it to robot-bazaar, etc etc.

pattacini:
>We can backport to 2021.05 if you like, but we would need a new tag in any case to properly propagate it to robot-bazaar, etc etc.

I think we're fine anyway 👍🏻

Issue #779, opened 2021-06-01
Test robotology-superbuild with zsh


traversaro:
The new default shell in macOS is zsh  since 10.15 (see https://scriptingosx.com/2019/06/moving-to-zsh/). Apparently, most users of the robotology-superbuild still do not use zsh, but we should nevertheless test it if it works fine, so we are not surprised when users try to use it.

Things to test:
* [x] Build everything from source, and source the `<install_prefix>/share/robotology-superbuild/setup.sh`  script of the superbuild, and check if everything is correct or if that script contains some bash-ism 
* [ ] Try to install binary packages via conda following the instructions in https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-forge.md . In theory for our packages we already generate setup scripts for zsh via [multisheller](https://github.com/mamba-org/multisheller), see https://github.com/robotology/robotology-superbuild/blob/545ec06c6915e0581dfaa2a09cc435f37ad5d614/conda/python/generate_conda_recipes_from_metametadata.py#L80 , but in practice that is completed untested.

nunoguedelha:
Ciao @traversaro , I guess that for testing you will first do:
- Replace the shebang `#!/bin/bash` by `#!/bin/zsh` or `#!/bin/sh`.
- Change the symbolic link stored in `/var/select/sh` to point the zsh shell (probably done by `chsh -s /bin/zsh`).
- Translate the ~/.bash_profile config to the zsh configuration files for zsh.

Please let me know if you need help for this.

traversaro:
Even without switching the default terminal of your system to `zsh`, I guess an easy check given an existing installation is to open a zsh terminal, and just try to source the `setup.sh` and check if the program start as usual or not. If you have already a robotology-superbuild installation available, that should be easy to do. By the way, I just realized this should be relatively easy do do even on Linux, just by installing zsh.

traversaro:
I tested on a Ubuntu 18.04 docker image zsh (`sudo apt install zsh`) and the setup.sh is correctly sourced, including features as checking if a file exits with `-f` and the syntax to append a value only of the variable is not defined.

lrapetti:
> * Build everything from source, and source the `<install_prefix>/share/robotology-superbuild/setup.sh`  script of the superbuild, and check if everything is correct or if that script contains some bash-ism

I did build everything from source again, and I have blindly copied my `.bash_profile` into the `.zshrc`. The `setup.sh` seems to be working fine. 

[...]

PR #810, opened 2021-06-29
Fix support for zsh in conda binary activation scripts

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/779 . This PR should fix all the `pathappend:3: bad substitution` experienced by zsh users such as @lrapetti or @Giulero .

We temporary install `multisheller` from a fixed commit to include the required fix https://github.com/mamba-org/multisheller/pull/13 .


traversaro:
Please ignore the CI failures, there are usual network non-reproducible issues. The relevant CI job is https://github.com/robotology/robotology-superbuild/actions/runs/982744267, and that was able to produce conda packages.

traversaro:
Friendly ping @lrapetti @Giulero @nunoguedelha (I tagged you as zsh users), as usual there is no need to understand everything in depth in the PR, just checking if it make sense is probably enough.

lrapetti (approved):
Thanks @traversaro, sorry but I was AFK last week.
The idea overall seems ok, even if am not aware of some details.

nunoguedelha:
Sorry @traversaro , I haven't tried at all yet using `zsh` shell.

PR #825, opened 2021-07-14
Document required apt dependencies in machine readable format with apt.txt file

traversaro:
A trend that we noticed in the latest years was that the long list of apt dependencies of the superbuild tended to be duplicated in many places (Docker image recipes, documentation pages on how to install the superbuild). This clearly created problem anytime the list of dependencies changed, as all the hardcoded copies of the list were outdated and needed to be manually updated. 

To solve this problem, this PR does the following:
* The main apt dependencies of the superbuild (the one documented in the main readme as dependencies of the superbuild as an whole) are now stored in the `apt.txt` file in the root of the repo (convention copied by repo2docker, see https://repo2docker.readthedocs.io/en/latest/config_files.html#apt-txt-install-packages-with-apt-get). The rest of dependencies that are required only for specific options are instead still documented manually, but I noticed that in practice those dependencies are much less used. 
* Instead of the manual list of dependencies, now the README suggest to run the `sudo sh ./scripts/install_apt_dependencies.sh` script to install the dependencies, that just install via apt-get all the dependncies listed in the apt.txt file.

This change will be released first in Distro Release 2021.08, so unfortunatly it will be possible to use it only from that point on.


apt.txt
```diff
@@ -0,0 +1,39 @@
...
+qml-module-qt-labs-settings
+qml-module-qtmultimedia
+qml-module-qtquick-controls
+qml-module-qtquick-dialogs
+qml-module-qtquick-window2
+qml-module-qtquick2
+qtbase5-dev
+qtdeclarative5-dev
+qtmultimedia5-dev
+swig libmatio-dev
```
S-Dafarra, line 35:
```suggestion
swig
libmatio-dev
```

S-Dafarra (approved):

Nicogene (approved):

traversaro:
fyi @pattacini

pattacini:
Nice feature!

PR #833, opened 2021-07-19
conda CI: Remove pinning to old libsolv version

traversaro:
All the problem should be solved by recent mamba versions, see: 
* https://github.com/robotology/robotology-superbuild/issues/785
* https://github.com/mamba-org/mamba/issues/986


traversaro:
Indeed, it seems that everything is installed correctly.

Nicogene (approved):

PR #847, opened 2021-08-05
Migrate conda documentation to use mamba instead of conda

traversaro:
This is a result of a feedback that came from our first training on conda: the time taked by conda install was incompatible with quickly explaing how conda environment works, so better to officially switch to mamba to ensure that the users have a smooth and fast user experience.


traversaro:
Please ignore the CI as this is a documentation-only change.

diegoferigo (approved):

PR #849, opened 2021-08-05
[Do Not Merge] Disable ROBOTOLOGY_ENABLE_ROBOT_TESTING option on Conda\Linux

traversaro:
To debug if other projects have the same prolem that we have in https://github.com/robotology/robotology-superbuild/issues/848 with blocktest.


traversaro:
Indeed, with `ROBOTOLOGY_ENABLE_ROBOT_TESTING` set to OFF the problem does not appear, so probably diagnostic-daemon is not a problem.

Issue #842, opened 2021-07-30
Stop supporting Homebrew in robotology-superbuild Continuous Integration and direct macOS users to conda?


traversaro:
Homebrew CI is extremely fragile, for several reasons. The main one, is that there is no easy way in Homebrew to be able to install the same package you installed a few days before, basically making really difficult to achieve reproducibility, that is quite critical in scientific research. 

To give an idea, this are just the issue related to Homebrew CI failures just in the last few months: 
* https://github.com/robotology/robotology-superbuild/issues/841
* https://github.com/robotology/robotology-superbuild/issues/826
* https://github.com/robotology/robotology-superbuild/issues/789
* https://github.com/robotology/robotology-superbuild/issues/704
* https://github.com/robotology/robotology-superbuild/issues/693
* https://github.com/robotology/robotology-superbuild/issues/571
* https://github.com/robotology/robotology-superbuild/issues/568
* https://github.com/robotology/robotology-superbuild/issues/565
* https://github.com/robotology/robotology-superbuild/issues/490
* https://github.com/robotology/robotology-superbuild/issues/476
* https://github.com/robotology/robotology-superbuild/issues/440
* https://github.com/robotology/robotology-superbuild/issues/879
* https://github.com/robotology/robotology-superbuild/pull/899#issuecomment-934541671, https://github.com/actions/virtual-environments/issues/4228
* https://github.com/robotology/robotology-superbuild/issues/919
* https://github.com/robotology/robotology-superbuild/issues/1017

Continuously dealing with these CI failures is quite demanding. At the same time, during the last year I think that `conda-forge` and the related channels (`robostack` and `robotology`) reached a level of maturity for which they are able to provide to macOS users easily installable and reproducible binaries (see . 

At this point I think a reasonable plan for the future, unless someone step in to volunteer mantainance of the Homebrew CI, is to **stop supporting the use of Homebrew in the robotology-superbuild CI and docs**, and direct macOS users to use conda instead. 

To clarify, this does not mean that single software projects part of the robotology organization cannot support Homebrew: anyone can support Homebrew, simply there will be no CI at the robotology-superbuild level to check that all packages are able to compile w.r.t. to compatible versions or branches of each other.

Ideally we would like all regular robotology-superbuild macOS user to be proficient with conda before that day, but I guess that unless there is a reason for doing so, people will not go out of their Homebrew-using comfort zone, and I know this process take time, so better to start off.

[...]

PR #852, opened 2021-08-06
Revise main README to mention conda Binary and Source installation

traversaro:
This PR revise the robotology-superbuild README with the following changes: 
* It provides pointers on how to use conda to either install binaries of the robotology packages via conda, or how to compile the superbuild using conda-forge dependencies. This would ensure that more users outside IIT are exposed to this new tools, that we have been using internally at IIT for some months now.
* It moves the documentation of the Homebrew and vpckg-based installation methods to a separate `doc/deprecated-installation-methods.md` page, to try to clarify for users that those methods are not our preferred way of installing software on macOS and Windows, see https://github.com/robotology/robotology-superbuild/issues/842 .


traversaro:
The README probably would be benefic for a general overhaul, but I would prefer first to get this change in, instead of trying to both introduce conda in the README, and change its structure at the same time.

diegoferigo (approved):

Nicogene (approved):

Issue #854, opened 2021-08-19
Debian Sid CI is failing (August 2021)


traversaro:
Error:
~~~
2021-08-19T02:10:12.5623285Z Reading package lists...
2021-08-19T02:10:12.7064251Z Building dependency tree...
2021-08-19T02:10:12.7065126Z Reading state information...
2021-08-19T02:10:12.7127284Z E: Unable to locate package libdc1394-22-dev
2021-08-19T02:10:12.7214286Z ##[error]Process completed with exit code 123.
~~~

Not a big problem now, but it could be a problem for Ubuntu 22.04 .

traversaro:
It seems that indeed `libdc1394-22-dev` was renamed to  `libdc1394-dev`, see: 
* https://packages.debian.org/search?keywords=libdc1394
* https://packages.ubuntu.com/search?keywords=libdc1394

Unfortunatly, while Ubuntu 20.04 and the last Debian stable (`bullseye`) have already `libdc1394-dev` and also a temporary `libdc1394-22-dev` package, Ubuntu 18.04 and Debian stretch still only have `libdc1394-22-dev`, so we will need some conditional logic/documentation somewhere.

traversaro:
Note that the latest pc104 image is based on Debian Buster: https://icub-tech-iit.github.io/documentation/icub_operating_systems/pc104/icub-live/ .

traversaro:
Unfortunatly, installing `libdc1394-dev` on Ubuntu 20.04 seems to uninstall some parts of opencv for some reason:
~~~
The following additional packages will be installed:
  libdc1394-25
The following packages will be REMOVED:
  libdc1394-22-dev libopencv-calib3d-dev libopencv-contrib-dev libopencv-dev libopencv-features2d-dev libopencv-highgui-dev libopencv-objdetect-dev
  libopencv-stitching-dev libopencv-videostab-dev
The following NEW packages will be installed:
  libdc1394-25 libdc1394-dev
~~~

So better to continue to use `libdc1394-22-dev` on all distros that we can.

PR #855, opened 2021-08-25
Handle different names of libdc1394 package in different distributions

traversaro:
This fixes https://github.com/robotology/robotology-superbuild/issues/854 . I took the occasion to also move in `install_apt_dependencies.sh` the installation of nlohmman-json library.

### DIFF
--- a/scripts/install_apt_dependencies.sh
+++ b/scripts/install_apt_dependencies.sh
@@ -3,6 +3,8 @@
 # Install dependencies of robotology-superbuild 
 # using apt on Ubuntu or Debian 
 
+lsb_release -a
+
 # Get location of the script
 SCRIPT_DIR=$(cd "$(dirname "$BASH_SOURCE")"; cd -P "$(dirname "$(readlink "$BASH_SOURCE" || echo .)")"; pwd)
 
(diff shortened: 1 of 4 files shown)


scripts/install_apt_dependencies.sh
```diff
@@ -7,3 +7,28 @@
...
+  apt-get install -y libdc1394-22-dev
+fi
+
+# Handle nlohmann-json3-dev dependency
+# Remove once Ubuntu 18.04 compatibility is dropped
+if [[ ("$ROBSUP_DISTRO_NAME" == "Ubuntu" && "$ROBSUP_DISTRO_VERSION" == "18.04") ]]
+then
+  :
+else
+  apt-get install -y nlohmann-json3-dev
```
Nicogene, line 33: Which version of the library is it installed? Because now `yarp-telemetry` needs the `v3.10.0`

traversaro: The specific version depends on the specific ubuntu version, see https://repology.org/project/nlohmann-json/versions . However, at the moment not even Debian Unstable (sid) has 3.10.0 . I think we can continue to use the FetchContent-provided nlohmann-json in yarp-telemetry for some time now.

traversaro:
The CI is now green.

Nicogene (approved):

PR #844, opened 2021-08-03
Fix non-deterministic compilation failure in walking-teleoperation when ROBOTOLOGY_ENABLE_HUMAN_DYNAMICS and ROBOTOLOGY_ENABLE_TELEOPERATION are enabled

traversaro:
Fix https://github.com/robotology/walking-teleoperation/issues/65 .

The problem was two fold: 
* `HumanDynamicsEstimation`  was called `human-dynamics-estimation`, even if its CMake config file is   `HumanDynamicsEstimation` 
* The dependency of `walking-teleoperation`  on `HumanDynamicsEstimation`  was not part of the configuration files, so there was nothing ensuring that `walking-teleoperation`  was built after `HumanDynamicsEstimation`

PR #858, opened 2021-08-30
Remove use of devel branch for walking-teleoperation stable branch

traversaro:
The need for it emerged in https://github.com/robotology/robotology-superbuild/pull/844#issuecomment-893293323 but was fixed by https://github.com/robotology/walking-teleoperation/pull/67 .


lrapetti (approved):
Looks good

PR #873, opened 2021-09-02
Always install Gazebo 11 in apt CI

traversaro:


traversaro:
The only two failures are related to https://github.com/robotology/robotology-superbuild/issues/872#issuecomment-911703973 , so not related to this PR. I think this is ready to be reviewed and merged.

cmake/ProjectsTagsUnstable.cmake
```diff
@@ -33,4 +33,4 @@ set_tag(OsqpEigen_TAG master)
 set_tag(YARP_telemetry_TAG master)
 set_tag(gym-ignition_TAG master)
 set_tag(walking-teleoperation_TAG devel)
-
+et_tag(speech_TAG fix/40)
```
diegoferigo, line 36:
```suggestion
set_tag(speech_TAG fix/40)
```

diegoferigo (approved):

cmake/ProjectsTagsUnstable.cmake
traversaro: Thanks, this is not necessary anymore after https://github.com/robotology/speech/pull/41 .

Nicogene (approved):

traversaro:
Not sure why we also have a libdc1394-related failure in sid, however I try to fix also that one.

PR #874, opened 2021-09-03
Fix link in README's TOC

traversaro:


traversaro:
@sgiraz this was something we encountered when checking out the docs

Nicogene (approved):

Issue #30, opened 2018-01-30
Document the ROBOTOLOGY_USES_LUA option


traversaro:
Once it is documented, we can remove the `mark_as_advanced`.

traversaro:
Original code in the README removed in https://github.com/robotology/robotology-superbuild/pull/31 :

| `ROBOTOLOGY_USES_LUA`  | Include software and plugins that depend on the [Lua scripting language](https://www.lua.org/). | `OFF` | [Documentation on Lua dependency.](#lua) | 


## Lua 
Support for this dependency is enabled by the `ROBOTOLOGY_USES_LUA` CMake option.

### Configuration
**TODO**

### Check the installation
**TODO**

traversaro:
Another existing problem (cc @xEnVrE) is that at the moment the `setup.sh` ( https://github.com/robotology/robotology-superbuild/blob/master/cmake/template/setup.sh.in ) and all related files in Windows (see https://github.com/robotology/robotology-superbuild/issues/147) do not set the `LUA_CPATH`  env variable, so you will need to set it manually.

traversaro:
@Nicogene probably experienced a regression as this option is not tested in CI:
~~~
In file included from /Users/runner/work/icub-tests/icub-tests/yarp/src/carriers/portmonitor_carrier/MonitorBinding.cpp:14: 

 

/Users/runner/work/icub-tests/icub-tests/yarp/src/carriers/portmonitor_carrier/lua/MonitorLua.h:16:10: fatal error: 'swigluarun.h' file not found 

 

#include "swigluarun.h" 

 

 ^~~~~~~~~~~~~~ 

 

1 error generated. 

 

make[2]: *** [src/carriers/portmonitor_carrier/CMakeFiles/yarp_portmonitor.dir/MonitorBinding.cpp.o] Error 1 

 

make[1]: *** [src/carriers/portmonitor_carrier/CMakeFiles/yarp_portmonitor.dir/all] Error 2 

 
make: *** [all] Error 2
~~~

PR #876, opened 2021-09-03
Update apt.txt wrt `lua`

pattacini:
I remember back in time that the IOL demo needed also the plain `lua` package: see the [YARP doc](https://github.com/robotology/yarp/commit/fa5035fad77d2668ff1cbd82372a3813d4e4e2d4) I updated in this respect.
It's maybe also the right time to upgrade to version `5.2`.

### DIFF
--- a/apt.txt
+++ b/apt.txt
@@ -13,7 +13,8 @@ libedit-dev
 libeigen3-dev
 libgsl0-dev
 libjpeg-dev
-liblua5.1-dev
+liblua5.2-dev
+lua5.2
 libode-dev
 libopencv-dev
 libsdl1.2-dev


pattacini:
Given https://github.com/icub-training/icub-training.github.io/pull/8#issuecomment-912702156, we might want to remove the `dev` part then.

traversaro:
In theory, that dependency is only necessary if `ROBOTOLOGY_USES_LUA`  (see https://github.com/robotology/robotology-superbuild/issues/30) is enabled. However, at least for small dependency it is indeed just easier to have them in the main `apt.txt`  file, so they are already there if necessary, as the downside is quite limited (just a few moments of installation time and some disk space used).

traversaro:
In particular, lua5.2 seems to be availale for all the supported Debian ( https://packages.debian.org/search?keywords=lua5 ) and Ubuntu versions (https://packages.ubuntu.com/search?keywords=lua5 ). 

Note that even lua5.3 and lua5.4 are available, but if you used/tested lua5.2 let's stick to it.

traversaro (approved):

PR #870, opened 2021-09-01
Move YARP_TAG of ProjectsTagsStable to yarp-3.5

traversaro:
This means that users that use the `Stable` setting of the robotology-superbuild (that is the default) will start getting the new yarp 3.5 instead of yarp 3.4 . This may be reveal new problems as several changes are present, so please report any new issue.

Issue #879, opened 2021-09-06
 macOS Homebrew CI failure 2021/09/06 


traversaro:
~~~
==> Pouring python@3.9--3.9.7.catalina.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3
Target /usr/local/bin/2to3
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3'

To force the link and overwrite all conflicting files:
  brew link --overwrite python@3.9

To list all files that would be deleted:
  brew link --overwrite --dry-run python@3.9

Possible conflicting files are:
/usr/local/bin/2to3 -> /Library/Frameworks/Python.framework/Versions/2.7/bin/2to3
~~~

traversaro:
Related issue tracker: https://github.com/actions/virtual-environments/issues/4020 .

traversaro:
Apparently the issue is now fixed: https://github.com/robotology/robotology-superbuild/actions/runs/1215640773 .

PR #881, opened 2021-09-06
Remove use of devel branch of gazebo-yarp-plugins in ProjectsTagsStable

traversaro:
This was added in https://github.com/robotology/robotology-superbuild/pull/870, but now that both gazebo-yarp-plugins master and devel are compatible with YARP v3.5.0, we can go back to use master in stable branches.


traversaro:
Please ignore build failures in macOS/Homebrew as those are related to https://github.com/robotology/robotology-superbuild/issues/879 .

diegoferigo (approved):

Issue #871, opened 2021-09-01
Incorrect QT path configuration when installing robotology-superbuild from source on Windows with conda


traversaro:
Thanks to @sgiraz tests, it seems that when following the steps in https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-forge.md#run-software-installed, to actually run graphical users interfaces such as FirmwareUpdater, the `QT_QPA_PLATFORM_PLUGIN_PATH` env variable should be defined and point to `%CONDA_PREFIX%\Library\plugins\platforms` . We need to understand if this is a bug in https://github.com/conda-forge/qt-feedstock or something that we need to address on our side.

sgiraz:
I would notify another potential problem that could be related to the `conda-Qt` installation.
Once the `yarpmotorgui` is open, the graphics of some button does not appear, despite the buttons and their tooltip exist.

![InkedMicrosoftTeams-image (1)_LI](https://user-images.githubusercontent.com/18039045/131856548-a6c1286d-4a62-42a8-a28e-8dc7922b9ac9.jpg)

traversaro:
Just to double check, I launched `yarpmotorgui` installed from YARP conda binaries (so without compiling from source the superbuild) and indeed there is no need to of setting `QT_QPA_PLATFORM_PLUGIN_PATH`, and the icons can be found:
![yarpmotorgui](https://user-images.githubusercontent.com/1857049/132015304-3e166fad-1385-4eba-a0aa-f18b8b3923f3.PNG)

sgiraz:
<details>
  <summary>Here my `conda-list`:</summary>

``` console
# Name                    Version                   Build  Channel
ace                       7.0.3                h0e60522_0    conda-forge
asio                      1.14.1               h0e60522_1    conda-forge
boost                     1.74.0           py39hefe7e4c_3    conda-forge
boost-cpp                 1.74.0               h5b4e17d_4    conda-forge
bzip2                     1.0.8                h8ffe710_4    conda-forge
c-compiler                1.2.0                h8ffe710_0    conda-forge
ca-certificates           2021.5.30            h5b45459_0    conda-forge
cairo                     1.16.0            hb19e0ff_1008    conda-forge
cfitsio                   3.470                h0af3d06_7    conda-forge
clang_variant             1.0                     default    conda-forge
clangdev                  5.0.0             h5aa907a_1012    conda-forge
cmake                     3.21.2               h39d44d4_0    conda-forge
[...]

PR #883, opened 2021-09-07
Add workaround for non-relocatable qt used in conda-forge

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/871 . 

Based on my understanding, the problem is that the binary created by conda is not relocatable, so it contains the wrong paths (in particular, the one of the build machine, not the one of the machine in which it is installed).

In theory, this is a problem that should be solved at https://github.com/conda-forge/qt-feedstock . However, there is already a lot going on on that repo, so it is not even sure if a fix would be merged. Furthermore, updating to Qt 5.15 would solve the problem (thanks to https://www.qt.io/blog/qt-is-relocatable), so it is better to simply wait for 5.15 to arrive in qt-feedstock. In the meanwhile, a simple workaround is to have the superbuild copy the `qt.conf` in the right directory if it detect that is is being compiled on Windows in a conda environment. This is far from a perfect solution (it does not work if you compile projects on their own outside of the robotology-superbuild) but it will work in the short term.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,13 @@ if(WIN32)
                               ${YCM_EP_INSTALL_DIR}/share/${PROJECT_NAME}/removePathsFromUserEnvVariables.ps1 @ONLY)
 endif()
 
+# Install qt.conf on Windows on conda as a workaround for the non-relocatable qt version used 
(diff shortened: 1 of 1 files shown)


traversaro:
fyi @sgiraz

Nicogene (approved):

Issue #811, opened 2021-06-29
Problem in activating environment in Powershell in Windows when powershell profile is stored in OneDrive


traversaro:
@ste93 is having problems in activating arbitrary conda environment in powershell (after following the docs in https://github.com/robotology/robotology-superbuild/blob/master/doc/install-miniforge.md), while in command prompt everything is working. In particular `conda activate <nameenv>`  does not seem to do anything.

This is the output of his `conda init`:
~~~
PS C:\Users\sbernagozzi> conda init
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\Scripts\conda.exe
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\Scripts\conda-env.exe
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\Scripts\conda-script.py
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\Scripts\conda-env-script.py
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\condabin\conda.bat
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\Library\bin\conda.bat
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\condabin\_conda_activate.bat
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\condabin\rename_tmp.bat
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\condabin\conda_auto_activate.bat
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\condabin\conda_hook.bat
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\Scripts\activate.bat
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\condabin\activate.bat
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\condabin\deactivate.bat
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\Scripts\activate
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\Scripts\deactivate
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\etc\profile.d\conda.sh
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\etc\fish\conf.d\conda.fish
no change     C:\Users\sbernagozzi\AppData\Local\miniforge3\shell\condabin\Conda.psm1
[...]

PR #885, opened 2021-09-10
Document Powershell pitfalls and possible problems

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/811


ste93 (approved):

ste93:
Perfect, so new user can see the problem and fix it quickly ;) Thanks @traversaro

PR #887, opened 2021-09-16
Create 2021.08.1.yaml

Nicogene:
Preparing the upcoming distro release


pattacini (approved):

traversaro:
If we want to tag after the PR, probably you also want to change https://github.com/robotology/robotology-superbuild/blob/releases/2021.08/cmake/RobotologySuperbuildOptions.cmake#L72 .

traversaro:
> 
> 
> If we want to tag after the PR, probably you also want to change https://github.com/robotology/robotology-superbuild/blob/releases/2021.08/cmake/RobotologySuperbuildOptions.cmake#L72 .

This is also useful to actuall run the tests on 2021.08.1 tags.

traversaro:
Indeed, in this format the superbuild is not compiling. Can you use the `.yaml`  extension for consistency?

PR #891, opened 2021-09-20
Clarify in documentation that Powershell does not fully work for conda-based installation

traversaro:
This should be more clear w.r.t. to previously available documentation on the fact that as long as we don't fix some issues, it is better to avoid use powershell.

fyi @S-Dafarra @ste93


S-Dafarra (approved):

ste93:
perfect, thank you @traversaro

Issue #923, opened 2021-11-19
Compiling error with dependencies from conda-forge on Manjaro 


HosameldinMohamed:
I am trying to install the superbuild on Manjaro. But I get an error:

Following [the documentation](https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-forge.md#source-installation). Here's what I have tried.

```
mamba create -n robTest
conda activate robTest
mamba install -c conda-forge cmake compilers make ninja pkg-config
mamba install -c conda-forge ace asio assimp boost eigen gazebo glew glfw gsl ipopt irrlicht libjpeg-turbo libmatio libode libxml2 nlohmann_json opencv pkg-config portaudio qt sdl sdl2 sqlite tinyxml spdlog
mamba install -c conda-forge bash-completion expat-cos6-x86_64 freeglut libdc1394 libselinux-cos6-x86_64 libxau-cos6-x86_64 libxcb-cos6-x86_64 libxdamage-cos6-x86_64 libxext-cos6-x86_64 libxfixes-cos6-x86_64 libxxf86vm-cos6-x86_64 mesalib mesa-libgl-cos6-x86_64
```

Then cloned and build `robotology-superbuild`. I only disabled the option `ROBOTOLOGY_USES_GAZEBO` in cmake.

And here's the error I am getting:
```
[ 28%] Generating code from ../../../extern/ros/geometry2/tf2_msgs/srv/FrameGraph.srv
/home/hosameldin/Github/robotology-superbuild/src/YARP/src/carriers/portmonitor_carrier/lua/MonitorLua.cpp:9:10: fatal error: lua.hpp: No such file or directory
    9 | #include <lua.hpp>
      |          ^~~~~~~~~
compilation terminated.
[ 28%] Building CXX object src/libYARP_serversql/src/CMakeFiles/YARP_serversql.dir/yarp/serversql/impl/ConnectManager.cpp.o
make[5]: *** [src/carriers/portmonitor_carrier/CMakeFiles/yarp_portmonitor.dir/build.make:151: src/carriers/portmonitor_carrier/CMakeFiles/yarp_portmonitor.dir/lua/MonitorLua.cpp.o] Error 1
make[5]: *** Waiting for unfinished jobs....
[ 28%] Building CXX object src/libYARP_serversql/src/CMakeFiles/YARP_serversql.dir/yarp/serversql/impl/ConnectThread.cpp.o
[...]

PR #931, opened 2021-11-26
Install lua as dependency in conda-forge-based installation

traversaro:
This is done to mitigate cross-talking problems when compiling YARP in a conda-environment in Arch-based distributions, see https://github.com/robotology/robotology-superbuild/issues/923 .

This is just a mitigation, as the issue is either in the conda-packaged CMake that is able to find system libraries or in the YARP build system, but given that lua is a relative small dependency is not a problematic mitigation.


HosameldinMohamed (approved):
Thanks @traversaro !

Issue #927, opened 2021-11-23
Conda Package Generation failure Late November 2021


traversaro:
Error:
~~~
Traceback (most recent call last):
  File "/usr/share/miniconda/envs/test/bin/conda-mambabuild", line 7, in <module>
    from boa.cli.mambabuild import main
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/boa/cli/mambabuild.py", line 17, in <module>
    from boa.core.solver import MambaSolver
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/boa/core/solver.py", line 22, in <module>
    from mamba import mamba_api
ImportError: cannot import name 'mamba_api' from 'mamba' (/usr/share/miniconda/envs/test/lib/python3.8/site-packages/mamba/__init__.py
~~~

Diff between latest working version:
~~~diff
--- <unnamed>
+++ <unnamed>
@@ -5,7 +5,7 @@
 backports                 1.0                        py_2    conda-forge
 backports.functools_lru_cache 1.6.4              pyhd8ed1ab_0    conda-forge
 beautifulsoup4            4.10.0             pyha770c72_0    conda-forge
-boa                       0.7.1              pyha770c72_0    conda-forge
+boa                       0.5.0            py38h578d9bd_0    conda-forge
 brotlipy                  0.7.0           py38h497a2fe_1003    conda-forge
 bzip2                     1.0.8                h7f98852_4    conda-forge
 c-ares                    1.18.1               h7f98852_0    conda-forge
@@ -17,19 +17,20 @@
 clyent                    1.2.2                      py_1    conda-forge
 colorama                  0.4.4              pyh9f0ad1d_0    conda-forge
 commonmark                0.9.1                      py_0    conda-forge
-conda                     4.10.3           py38h578d9bd_3    conda-forge
-conda-build               3.21.6           py38h578d9bd_0    conda-forge
-conda-forge-pinning       2021.11.15.20.12.07      hd8ed1ab_0    conda-forge
+conda                     4.10.3           py38h578d9bd_4    conda-forge
+conda-build               3.21.6           py38h578d9bd_1    conda-forge
+conda-forge-pinning       2021.11.22.15.24.15      hd8ed1ab_0    conda-forge
 conda-package-handling    1.7.3            py38h497a2fe_1    conda-forge
 cryptography              35.0.0           py38h3e25421_2    conda-forge
[...]

PR #935, opened 2021-12-01
Fix documentation of ROBOTOLOGY_USES_ESDCAN when using conda and enable ROBOTOLOGY_USES_ESDCAN  in conda binary packages

traversaro:
Furthermore, enable ROBOTOLOGY_USES_ESDCAN when compiling conda binaries, so that people that install `icub-main` from binaries already have it enabled.

At the moment, this PR is blocked on https://github.com/robotology/robotology-conda-binary-repackaging-recipes/pull/1 . Once that PR is merged, we can merge this PR. If we are able to merge this in time for the release 2021.11, this is probably worth a mention in the release notes, as several Windows users are interested in using the `ecan` device driver for dumping data from `CAN` devices on Windows. fyi @Nicogene 

Fix https://github.com/robotology/robotology-superbuild/issues/934 .

PR #937, opened 2021-12-02
Fix conda package generation by pinning boa and mamba to known working versions

traversaro:
* `boa` and `mamba` were quite unstable lately, so pin them to safe version at least for the near future.
* There are strange internal compiler errors when building gazebo-yarp-plugins, so let's build stick to build it against gazebo 11.8 for which this problem do not emerge.
* Add the channel `robotology` when icub-main is built so that the `esdcan` dependency is correctly found (see https://github.com/robotology/robotology-superbuild/pull/935)

Fix https://github.com/robotology/robotology-superbuild/issues/927 .


traversaro:
This is now working, see: https://github.com/robotology/robotology-superbuild/actions/runs/1539056039 .

Nicogene (approved):

PR #942, opened 2021-12-06
generate-conda-packages: For icub-main switch to conda build and use strict channel_priority 

traversaro:
Improvements to complete : 
* it seems that `conda mambabuild` does not support correctly the `-c` option to specify the channels
* it seems that unless we set `--set channel_priority strict` for some reason when `icub-main` is built the `yarp` from robotology is found (instead of the one from `local`, i.e. that was previously built). 
* The `--skip-existing` option does not work as expected, so the `icub-main` configuration fails even if `icub-main` is already built. So we remove all the recipe directories manually, and then we build all the rest of recipes without `--skip-existing` option.
 
With these three fixes, icub-main with esd support is build correctly on Windows, completing the PR https://github.com/robotology/robotology-superbuild/pull/935 .

### DIFF
--- a/.github/workflows/generate-conda-packages.yaml
+++ b/.github/workflows/generate-conda-packages.yaml
@@ -112,28 +112,50 @@ jobs:
             # See https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#creating-conda-build-variant-config-files
             # We manually specify the build order as conda build is too slow, and conda mambabuild does not support correctly multiple recipes
             # see https://github.com/mamba-org/boa/issues/117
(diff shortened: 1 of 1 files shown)


traversaro:
See https://github.com/robotology/robotology-superbuild/actions/runs/1548511010 for a successul run of `generate-conda-packages` on this branch.

diegoferigo (approved):

traversaro:
> See https://github.com/robotology/robotology-superbuild/actions/runs/1548511010 for a successul run of `generate-conda-packages` on this branch.

I was too optimistic, a real successful run is https://github.com/robotology/robotology-superbuild/runs/4443869589 .

PR #954, opened 2021-12-14
Modify release workflow to account for new numbering scheme

traversaro:
After the change described in https://github.com/icub-tech-iit/documentation/pull/172, it is necessary to change the `release` workflow, otherwise the branches will be created with the name `releases/YYYY.MM.0` instead of the correct `releases/YYYY.MM`, as it happened for example with the 2021.11 release: https://github.com/robotology/robotology-superbuild/tree/releases/2021.11.0 .
For the `2021.11` I will manually modify the script, but this modification are there to avoid doing this error again in the future.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,19 +3,19 @@ name: Create a new release
 on:
   workflow_dispatch:
     inputs:
-      superbuild_tag:
-        description: 'The new robotology-superbuild tag(without the "v")'
+      superbuild_year_month_prefix:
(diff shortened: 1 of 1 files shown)


traversaro:
> Shall we instead reaquire a tag that is YYYY.MM.P without adding 0 at the end? It may be useful for deploying a release that it not the first one (the one with the patch number 0)

For a release that is not a .0, the stuff to do are quite different (you do not create the branch, you do not set Custom) so I was thinking to have a separate action that is "release-patch"  or similar, that take again in input the `YYYY.MM`, and just automatically computes the P required for the next `YYYY.MM.P` to create.

Nicogene (approved):

Issue #955, opened 2021-12-14
Conda compilation from source not compatible with macOS SDK in Xcode 13


traversaro:
Since yesterday, the CI of conda on macos-latest failed with:
~~~
2021-12-14T02:36:03.7886020Z [159/1268] Building CXX object src/libYARP_os/src/CMakeFiles/YARP_os.dir/yarp/os/impl/macos/MacOSAPI.mm.o
2021-12-14T02:36:03.7985980Z FAILED: src/libYARP_os/src/CMakeFiles/YARP_os.dir/yarp/os/impl/macos/MacOSAPI.mm.o 
[...]

PR #957, opened 2021-12-17
Workaround problem of multiple append in the same if in conda activation scripts in command prompt

traversaro:
Fix https://github.com/robotology/wb-toolbox/issues/228, using the same workaround that was already used in icub-models also in whole-body-controllers and wb-toolbox . 

For a brief explanation of what is going on, [multisheller](https://github.com/mamba-org/multisheller) is the project that we are using to generate "activation scripts" (i.e. the scripts that set the right enviromental variables when you type `conda activate <env>`) for our conda packages for all the terminal we are supporting (bash, zsh, command prompt). For a longer explanation, see last part of  https://github.com/robotology/robotology-superbuild/blob/0bf40cf7d2c05e130b6f7fc1c86b7a87c6fcd373/doc/developers-faqs.md#how-to-ensure-that-binary-packages-are-correctly-generated-for-a-new-package .

The problem is that multisheller on command prompt has a bug, that is described in https://github.com/mamba-org/multisheller/issues/9: when we are appending the same env var in the same if, in command prompt it is not set directly. As I do not know how to tackle the original issue, this PR just workaround the problem by ensuring that in wb-toolbox and whole-body-controllers a separate if is used for each call to append.

### DIFF
--- a/conda/multisheller/wb-toolbox_activate.msh
+++ b/conda/multisheller/wb-toolbox_activate.msh
@@ -1,6 +1,6 @@
-# For some reason appending two times the same variable inside the same if does not work in command prompt
+# For some reason, appending two times the same variable inside the same "if" statement does not work in command prompt.
 # See https://github.com/mamba-org/multisheller/issues/9
-# As a workaround, we move each append in its own if
+# As a workaround, we move each append in its own "if" statement.
 
 if_(is_set("COMSPEC")).then_([
 	sys.list_append("MATLABPATH", path.join(env("CONDA_PREFIX"), "Library\\share\\WBToolbox"))
--- a/conda/multisheller/wb-toolbox_activate.msh
+++ b/conda/multisheller/wb-toolbox_activate.msh
@@ -1,7 +1,15 @@
+# For some reason appending two times the same variable inside the same if does not work in command prompt
+# See https://github.com/mamba-org/multisheller/issues/9
+# As a workaround, we move each append in its own if
+
 if_(is_set("COMSPEC")).then_([
-	sys.list_append("MATLABPATH", path.join(env("CONDA_PREFIX"), "Library\\share\\WBToolbox")),
+	sys.list_append("MATLABPATH", path.join(env("CONDA_PREFIX"), "Library\\share\\WBToolbox"))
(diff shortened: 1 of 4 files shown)


traversaro:
cc @fabiodinatale as I am not sure you receive the notification for the review request.

traversaro:
Please ignore CI failures, there are related to the unrelated issues described in:
* https://github.com/robotology/robotology-superbuild/issues/955
* https://github.com/robotology/robotology-superbuild/issues/959

conda/multisheller/wb-toolbox_activate.msh
```diff
@@ -1,7 +1,15 @@
+# For some reason appending two times the same variable inside the same if does not work in command prompt
```
fabiodinatale, line 1: In my opinion, this sentence would be clearer if we change it like in the following:
```
For some reason, appending two times the same variable inside the same "if" statement does not work in command prompt.
```

fabiodinatale, line 1: In my opinion, this sentence would be clearer if we change it like in the following:
```
For some reason, appending two times the same variable inside the same "if" statement does not work in command prompt.
```

conda/multisheller/wb-toolbox_activate.msh
```diff
@@ -1,7 +1,15 @@
...
+# See https://github.com/mamba-org/multisheller/issues/9
+# As a workaround, we move each append in its own if
```
fabiodinatale, line 3: As before, I would change the sentence as follows:
```
As a workaround, we move each append in its own "if" statement.
```

fabiodinatale (approved):
The comments left on the first file hold also for the others since they are referring to the description of the problem.
On the code side, it looks ok to me.

traversaro:
Thanks, I implemented the suggested correction!

conda/multisheller/wb-toolbox_activate.msh
fabiodinatale: Ok!

traversaro:
I forgot to run the generate conda recipes on this job, and it seems it broke something.

PR #958, opened 2021-12-17
Switch conda CI to macos-10.15 workers

traversaro:
Workaround for https://github.com/robotology/robotology-superbuild/issues/955 . Indeed, switching to macos-10.15 works as Xcode 12 is used. However, the problem remains withw Xcode 13.


traversaro:
Merging directly without review to fix CI.

Issue #959, opened 2021-12-18
Unstable branches failure late December 2021 due to YARP's IRangefinder2D interface change


traversaro:
Since yesterday the unstable branches failed with error:
~~~
[...]

PR #961, opened 2021-12-18
Use YARP unstable checkout before IRangefinder2D interface changes

traversaro:
Workaround for https://github.com/robotology/robotology-superbuild/issues/959 .


traversaro:
Merge without review to fix CI.

PR #965, opened 2021-12-20
GazeboYARPPlugins: add dependency on dartsim when generating conda packages

traversaro:
~The use of 11.8 is generating the problem described in https://github.com/robotology/yarp.js/issues/28#issuecomment-997122408, let's try to remove this so the latest gazebo is used.~ 

As gazebo 11.9 still results in internal compiler error, we still require gazebo=11.8 in the build and for now we just add the dependency on dartsim so we get the correct run_exports dependency in gazebo-yarp-plugins.

### DIFF
--- a/cmake/BuildGazeboYARPPlugins.cmake
+++ b/cmake/BuildGazeboYARPPlugins.cmake
@@ -41,4 +41,6 @@ ycm_ep_helper(GazeboYARPPlugins TYPE GIT
                                         gazebo
                                 CMAKE_ARGS -DGAZEBO_YARP_PLUGINS_HAS_OPENCV:BOOL=ON)
 
-set(GazeboYARPPlugins_CONDA_DEPENDENCIES libopencv gazebo)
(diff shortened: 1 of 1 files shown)


traversaro:
Test `generate-conda-packages` run: https://github.com/robotology/robotology-superbuild/actions/runs/1601540116 .

traversaro:
Note that the `gazebo=11.8` constraint was added in https://github.com/robotology/robotology-superbuild/pull/937, hopefully is not necessary after https://github.com/robotology/robotology-superbuild/pull/964 .

traversaro:
> As gazebo 11.9 still results in internal compiler error, we still require gazebo=11.8 in the build and for now we just add the dependency on dartsim so we get the correct run_exports dependency in gazebo-yarp-plugins.

New test run to try this workaround: https://github.com/robotology/robotology-superbuild/actions/runs/1601909749 .

traversaro:
Merging as this is just a workaround to merge before the conda package generation scheduled for this night.

PR #968, opened 2021-12-20
Add workaround for gazebo-yarp-plugins internal compiler error by switching to clang

traversaro:
Instead of requiring an old version of Gazebo as done in https://github.com/robotology/robotology-superbuild/pull/937, we switch to use the clang++ compiler if gazebo-yarp-plugins is built. Not a complete fix, but better that use an old version of Gazebo that in the long run forces a old version of all the transitive dependencies of Gazebo.

Fix https://github.com/robotology/robotology-superbuild/issues/966 .


traversaro:
Generate conda package example build: https://github.com/robotology/robotology-superbuild/actions/runs/1604112386 .

traversaro:
Example build: https://github.com/robotology/robotology-superbuild/runs/4588477984 . It worked correctly!

traversaro:
@nunoguedelha this should remove the need for building gazebo-yarp-plugins against gazebo 11.8 .

Nicogene (approved):

diegoferigo (approved):

PR #970, opened 2021-12-20
Fix regression in generation of whole-body-controllers recipe introduced by PR 957

traversaro:
PR https://github.com/robotology/robotology-superbuild/pull/957 create a regression that broke the `generate-conda-recipes` job. This PR fixes it.


traversaro:
Merging without review to unblock the conda generation scheduled for this night.

PR #975, opened 2021-12-24
2021.11 release series: Bump version to 2021.11.1

traversaro:
First step towards a new release in the 2021.11 series (fyi @Nicogene this is the step I would like to automatize as discussed in https://github.com/robotology/robotology-superbuild/pull/954#issuecomment-993707057).

The file `2021.11.1.yaml` is copied from `2021.11.0.yaml`, we can bump the dependencies that we want to bump in a separate commit/PR.


Nicogene (approved):

traversaro:
Please only check Default CI jobs (the one that correspond to 2021.11.1 job), all other can be ignored.

Issue #974, opened 2021-12-23
CI Debian Testing failure end of december 2021


traversaro:
This seems to be an actual failure, and right in time for the Ubuntu 22.04 development:
~~~
2021-12-23T09:03:51.9872075Z CMake Error at src/solid-shapes/CMakeLists.txt:8 (add_library):
2021-12-23T09:03:51.9873667Z   Target "idyntree-solid-shapes" links to target "draco::draco_shared" but
2021-12-23T09:03:51.9874777Z   the target was not found.  Perhaps a find_package() call is missing for an
2021-12-23T09:03:51.9875394Z   IMPORTED target, or an ALIAS target is missing?
2021-12-23T09:03:51.9875736Z 
2021-12-23T09:03:51.9875925Z 
2021-12-23T09:03:51.9876424Z -- Generating done
~~~

traversaro:
This is a real issue.

`assimp` 5.1.3 contains these lines in the `assimpConfig.cmake` : 
~~~
# Create imported target assimp::assimp
add_library(assimp::assimp SHARED IMPORTED)

set_target_properties(assimp::assimp PROPERTIES
  INTERFACE_COMPILE_FEATURES "c_std_99"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "\$<\$<NOT:\$<CONFIG:DEBUG>>:/usr/lib/x86_64-linux-gnu/libz.so>;\$<\$<CONFIG:DEBUG>:/usr/lib/x86_64-linux-gnu/libz.so>;draco::draco_shared;minizip;/usr/lib/x86_64-linux-gnu/librt.so"
)
~~~

But no `find_dependency(draco)` is specified. 
 
I also checked in Debian sid, and `assimp` 5.1.4 contains this line in the `assimpConfig.cmake` : 
~~~
add_library(assimp::assimp SHARED IMPORTED)

set_target_properties(assimp::assimp PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include"
[...]

PR #976, opened 2021-12-24
Testing Debian Sid instead of Debian Testing

traversaro:
Not meant to be merged, but useful to debugging the issue from https://github.com/robotology/robotology-superbuild/issues/974 to see if it is still present in Debian sid that as of 2021/12/23 has assimp 5.1.4 while Testing has 5.1.3 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,7 +185,7 @@ jobs:
     - uses: actions/checkout@master
 
     - name: Upgrade apt packages Debian Testing and Sid [Debian Testing and Sid]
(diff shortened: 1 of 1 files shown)


traversaro:
Mhh, the CI fails before:
~~~
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 libignition-msgs-dev : Depends: libignition-msgs5-5 (= 5.1.0+dfsg-7+b1) but 5.8.1+ds-2 is to be installed
E: Unable to correct problems, you have held broken packages.
~~~

Issue #280, opened 2019-10-28
Use own repository in custom tag (ref. ROBOTOLOGY_PROJECT_TAGS)


valegagge:
I think the recently added feature about projects tag choice is very useful. I'm writing here to ask if it is possible to expand the "custom" tag by indicating the repository to use in addition to the tag in the custom file.

For example, my use case is the following:
I'd liker to crate a container with the superbuild as usual, but I'd like to use my icub-main fork where I developed some stuff.

I don't know if this is possible and how much effort its development requires, so please feel free to reject this proposal.

traversaro:
I did not document it because I am not really sure about all the implications, but all the arguments of the `ycm_ep_helper` can be overridden by CMake variables, see http://robotology.github.io/ycm/gh-pages/v0.10/manual/ycm-superbuild.7.html#overriding-parameters and https://github.com/robotology/ycm/blob/7ff014a958c3a7286644eab4324944de041ffcfb/modules/YCMEPHelper.cmake#L771 . For this reason, you can just add: 
~~~
set(ICUB_REPOSITORY valegagge/icub-main.git
~~~
and in theory (at least on a clean superbuild) you could specify the repo in place of the official one. However, I never tested this so I am not sure if it works or other problems arise.

traversaro:
Hi @valegagge , if the answer solves your issue, feel free to close this issue. If we want to document this feature, we can open a new issue to track this.

valegagge:
hi @traversaro , I think your answer is the solution, but I had no time to verify if it works. For this reason I don't close this issue yet. 
Feel free to close this issue and in case I'll get in trouble with your solution I'll reopen it.

traversaro:
Ok, let's wait for the verification if you plan to test it!

traversaro:
As I guess there is no plan to test this, I think we can close this issue @valegagge .

PR #980, opened 2021-12-30
Fix overriding of <package>_REPOSITORY and <package>_TAG when a distro yaml file is used

traversaro:
The procedure for using custom version of repository as described in https://github.com/robotology/robotology-superbuild/issues/280 was not working correctly when the version used were loaded from `.yaml`  file. So, the procedure was working correctly if the `Stable` or `Unstable` as those are loaded via custom .cmake files (see https://github.com/robotology/robotology-superbuild/blob/v2021.11.0/cmake/ProjectsTagsStable.cmake and https://github.com/robotology/robotology-superbuild/blob/v2021.11.0/cmake/ProjectsTagsUnstable.cmake) but not if the version were loaded by `.yaml`, as the `ycm_load_vcs_yaml_info` was overwriting the values specified via the CMake command line.

This PR should fix this.

### DIFF
--- a/cmake/YCMLoadVcsYamlInfo.cmake
+++ b/cmake/YCMLoadVcsYamlInfo.cmake
@@ -22,7 +22,10 @@
 # 
 #  If the VERBOSE option is passed, ycm_load_vcs_yaml_info will print all the variable
 #  that it sets.
-#  
+#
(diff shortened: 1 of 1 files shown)


traversaro:
@pattacini

traversaro:
CI seems happy, let's merge it.

PR #979, opened 2021-12-27
Pin Stable builds to v1.22.0 to workaround lack of YARP 3.6 release

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/978 .

PR #983, opened 2021-12-31
Update package versions in 2021.11.1 release

traversaro:
CI will probably fail as https://github.com/robotology/whole-body-estimators/pull/135 still needs to be merged and released, but better to open this PR so we can check if we are missing some release.

PR #984, opened 2022-01-03
Bump YARP and yarp-matlab-bindings in Stable branches to yarp-3.6

traversaro:
As `yarp` and `yarp-matlab-bindings` use a different branch for each minor release, we need to update manually them to ensure that Stable branches use YARP 3.6 .
As now we bump to yarp-3.6, I also removed the workaround added in https://github.com/robotology/robotology-superbuild/pull/979 .

### DIFF
--- a/cmake/ProjectsTagsStable.cmake
+++ b/cmake/ProjectsTagsStable.cmake
@@ -14,8 +14,7 @@ set_tag(casadi 3.5.5.3)
 
 # Robotology projects
 set_tag(YCM_TAG ycm-0.13)
-set_tag(YARP_TAG yarp-3.5)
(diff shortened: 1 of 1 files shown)


traversaro:
Probably this is also blocked by https://github.com/robotology/whole-body-estimators/pull/135 .

traversaro:
> Probably this is also blocked by [robotology/whole-body-estimators#135](https://github.com/robotology/whole-body-estimators/pull/135) .

It will also fail for https://github.com/robotology/robotology-superbuild/pull/983#issuecomment-1003957653 .

traversaro:
All failures are only related to `Unstable` branches, so I think we can merge this PR.

Nicogene (approved):

Issue #1002, opened 2022-01-19
Unable to find `i2c/smbus.h` (from conda-forge dependencies installation on Manjaro)


HosameldinMohamed:
Similar to #923 but when enabling `ROBOTOLOGY_ENABLE_ICUB_HEAD` ON, I was trying to install the superbuild on Manjaro. But I get an error:

Following [the documentation](https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-forge.md#source-installation). Here's what I have tried.

```
mamba create -n robTest
conda activate robTest
mamba install -c conda-forge cmake compilers make ninja pkg-config
mamba install -c conda-forge ace asio assimp boost eigen gazebo glew glfw gsl ipopt irrlicht libjpeg-turbo libmatio libode libxml2 nlohmann_json opencv pkg-config portaudio qt sdl sdl2 sqlite tinyxml spdlog lua
mamba install -c conda-forge bash-completion expat-cos6-x86_64 freeglut libdc1394 libselinux-cos6-x86_64 libxau-cos6-x86_64 libxcb-cos6-x86_64 libxdamage-cos6-x86_64 libxext-cos6-x86_64 libxfixes-cos6-x86_64 libxxf86vm-cos6-x86_64 mesalib mesa-libgl-cos6-x86_64
```

```
/home/hosameldin/Github/robotology-superbuild/src/YARP/src/devices/imuBosch_BNO055/imuBosch_BNO055.cpp:19:11: fatal error: i2c/smbus.h: No such file or directory
```

Even though `i2c/smbus.h` is available in the system files, in `/usr/include`. It's not clear to me when the compiler of Conda ignores system files and when it doesn't.

This is the package that provides the header file for instance https://archlinux.org/packages/community/x86_64/i2c-tools/

HosameldinMohamed:
@traversaro you mentioned something about not being able to find `I2C` package by Ubuntu/Debian.

```
ENABLE_yarpmod_imuBosch_BNO055 OFF - Dependencies unsatisfied: 'YARP_HAS_I2C' - imuBosch_BNO055-NOTFOUND
```

traversaro:
[...]

Issue #999, opened 2022-01-18
Missing libi2c-dev apt dependency? 


traversaro:
Some iCub models connect to the head's Bosch IMU bno055 IMU via i2c. To do so, they use YARP's device [`imuBosch_BNO055`](https://github.com/robotology/yarp/tree/v3.6.0/src/devices/imuBosch_BNO055) . To do so, on Debian systems you need to install the `libi2c-dev` dependency. However, this dependency is not listed anywhere.

traversaro:
@Nicogene did you ever had problems related to this?

Nicogene:
From what I remember no, but probably because in the [iCub OS images already contain it](https://icub-tech-iit.github.io/documentation/icub_operating_systems/download/)

PR #1003, opened 2022-01-21
Document installation of libi2c in Linux with apt and conda

traversaro:
Since a long time, we enabled the `YARP_USES_I2C` option on YARP. The option is automatically disabled if I2C is not found (and this is convenient for us on macOS and Windows), but the problem is that we never installed the required dependency. This PR fixes the problem by installing `libi2c-dev` from apt and `libi2c` from conda-forge.

Fix https://github.com/robotology/robotology-superbuild/issues/999 .
Fix https://github.com/robotology/robotology-superbuild/issues/1002 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
       run: |
         # Additional dependencies only useful on Linux
         # See https://github.com/robotology/robotology-superbuild/issues/477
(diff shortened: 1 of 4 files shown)


HosameldinMohamed (approved):
Thanks!

Nicogene (approved):

Issue #727, opened 2021-05-12
Add robotology-distro conda metapackage


traversaro:
Details to be inserted.

traversaro:
The overall idea of having a package such that is that we could do release of it corresponding to the releases of the superbuild. This package will not depend directly on any other package, but it would just list all the specific version required by a distro as `run_constrained`  entries, see https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained . 

In this way, if somebody wanted for example to install `yarp`, `icub-main`, `gazebo-yarp-plugins` and `icub-models` corresponding to the 2021.08 distro, they could do it easily by doing
~~~
conda install -c conda-forge -c robotology robotology-superbuild=2021.05 yarp icub-main  gazebo-yarp-plugins icub-models
~~~

The alternative (that we currently suggest in the docs added in https://github.com/robotology/robotology-superbuild/pull/852) is that the user need to manually inspect the tables in https://icub-tech-iit.github.io/documentation/sw_versioning_table/ to extract the version for each package that they need, so the previous install step will be instead:
~~~
conda install -c conda-forge -c robotology yarp=3.4.5 icub-main=1.20.0  gazebo-yarp-plugins=3.6.1 icub-models=1.20.0
~~~

But this is far from user friendly.

traversaro:
> `conda install -c conda-forge -c robotology robotology-superbuild=2021.05 yarp icub-main gazebo-yarp-plugins icub-models`

Reading this line makes me wonder if perhaps it is more clear if we call the metapackage `robotology-distro` instead of `robotology-superbuild`, as it effectively has nothing related to build anymore, and this may be more clear:
> `conda install -c conda-forge -c robotology robotology-distro=2021.05 yarp icub-main gazebo-yarp-plugins icub-models`

traversaro:
To simplify this, it has been agreed to name future releases that copy from `latest.releases.yaml` as `YYYY.MM.0`, while then patch releases as `YYYY.MM.1`, etc etc.

traversaro:
See https://github.com/icub-tech-iit/documentation/pull/172 .

traversaro:
I added some steps in this direction in https://github.com/traversaro/robotology-superbuild/tree/fix727 .

traversaro:
[...]

PR #1011, opened 2022-02-01
Bump qhull and CppAD to latest versions

traversaro:
This is part of the work on branch https://github.com/traversaro/robotology-superbuild/tree/fix727 that was done to solve https://github.com/robotology/robotology-superbuild/issues/727 . However, this part can be contributed on its own PR.

This PR bumps the qhull and CppAD version to track the latest version. In particular, this ensures that the versions used in the superbuild are compatible with the latest versions used in conda-forge:
* https://github.com/conda-forge/qhull-feedstock
* https://github.com/conda-forge/cppad-feedstock

Note that qhull has a bit of a complex versioning scheme, as it has releases like `8.0.2`  and `2020.2` . We switch to the `2020.2`-style tags for compatibility with most linux distros (see https://repology.org/project/qhull/versions) and with conda-forge.

As this are external packages, they are fixed to releases in `LatestReleases` and also for Stable and Unstable branches.

### DIFF
--- a/cmake/ProjectsTagsStable.cmake
+++ b/cmake/ProjectsTagsStable.cmake
@@ -8,8 +8,8 @@ endmacro()
 set_tag(osqp_TAG v0.6.2)
 set_tag(manif_REPOSITORY robotology-dependencies/manif.git)
 set_tag(manif_TAG 0.0.4.1)
-set_tag(qhull_TAG v8.0.2)
-set_tag(CppAD_TAG 20210000.8)
+set_tag(qhull_TAG 2020.2)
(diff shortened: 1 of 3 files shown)


GiulioRomualdi (approved):

GiulioRomualdi:
By the way, is this going to affect blf? There we require qhull v8

traversaro:
The 2020.2 tag was done later then the v8.0.* tags, so I think this should be ok: https://github.com/qhull/qhull/tags .

traversaro:
Anyhow, blf is compiling fine in the CI, so I think we can merge.

Issue #1010, opened 2022-02-01
Unstable Linux CI Failure Start February 2021


traversaro:
Probably just related to https://github.com/robotology/icub-firmware-shared/issues/55 .

traversaro:
> Probably just related to [robotology/icub-firmware-shared#55](https://github.com/robotology/icub-firmware-shared/issues/55) .

This was fixed by @marcoaccame in https://github.com/robotology/icub-firmware-shared/pull/56 . 
However, not all the builds are failing with error:
~~~
2022-02-03T03:17:56.6399805Z -- Found YARP: /home/runner/work/robotology-superbuild/robotology-superbuild/build/install/lib/cmake/YARP (found version "3.6.0+7-20220107.2+git38581de70")
2022-02-03T03:17:56.6400479Z CMake Error at cmake/WalkingControllersFindDependencies.cmake:152 (find_package):
2022-02-03T03:17:56.6401039Z   By not providing "FindBipedalLocomotionFramework.cmake" in
2022-02-03T03:17:56.6401545Z   CMAKE_MODULE_PATH this project has asked CMake to find a package
2022-02-03T03:17:56.6402330Z   configuration file provided by "BipedalLocomotionFramework", but CMake did
2022-02-03T03:17:56.6402740Z   not find one.
2022-02-03T03:17:56.6402960Z 
2022-02-03T03:17:56.6403206Z   Could not find a package configuration file provided by
2022-02-03T03:17:56.6403690Z   "BipedalLocomotionFramework" (requested version 0.6.0) with any of the
2022-02-03T03:17:56.6404234Z   following names:
2022-02-03T03:17:56.6404466Z 
2022-02-03T03:17:56.6404717Z     BipedalLocomotionFrameworkConfig.cmake
2022-02-03T03:17:56.6405277Z     bipedallocomotionframework-config.cmake
2022-02-03T03:17:56.6405974Z 
2022-02-03T03:17:56.6406294Z   Add the installation prefix of "BipedalLocomotionFramework" to
2022-02-03T03:17:56.6407349Z   CMAKE_PREFIX_PATH or set "BipedalLocomotionFramework_DIR" to a directory
2022-02-03T03:17:56.6407785Z   containing one of the above files.  If "BipedalLocomotionFramework"
2022-02-03T03:17:56.6408208Z   provides a separate development package or SDK, be sure it has been
2022-02-03T03:17:56.6408521Z   installed.
[...]

PR #1013, opened 2022-02-03
Add dependency of walking-controllers on bipedal-locomotion-framework

traversaro:
As `bipedal-locomotion-framework` is only compiled if `ROBOTOLOGY_ENABLE_DYNAMICS_FULL_DEPS` component is enabled, to make sure that everything works as expected we need to move `walking-controllers` under the `DYNAMICS_FULL_DEPS` component.

Fix https://github.com/robotology/robotology-superbuild/issues/1010 .


traversaro:
The dependency on BipedalLocomotionFramework was added in https://github.com/robotology/walking-controllers/pull/97, not sure why it started to fail just now.

traversaro:
> The dependency on BipedalLocomotionFramework was added in [robotology/walking-controllers#97](https://github.com/robotology/walking-controllers/pull/97), not sure why it started to fail just now.

The reason is that we were using the `master` branch of  `walking-controllers` so we were not testing  `walking-controllers`'s `devel`, see https://github.com/robotology/robotology-superbuild/pull/1014

lrapetti (approved):
Thanks for the fix @traversaro (I have just experienced this issue)

traversaro:
> Thanks for the fix @traversaro (I have just experienced this issue)

Thanks for the review, merging now.

PR #1014, opened 2022-02-03
When ROBOTOLOGY_PROJECT_TAGS is set to Unstable use devel walking-controllers branch

traversaro:
Before this PR the `master` branch was used, hiding problems such as the one catched (only when it reached the `Stable`/`master` branch) in https://github.com/robotology/robotology-superbuild/pull/1013 .


traversaro:
@GiulioRomualdi @S-Dafarra I opened this as I saw the master/devel pattern used in walking-controllers, if you want to switch to the single `master` branch also for walking-controllers please reply, so we do not merge this PR.

GiulioRomualdi:
Now also walking controller in master depends on blf.

S-Dafarra:
> @GiulioRomualdi @S-Dafarra I opened this as I saw the master/devel pattern used in walking-controllers, if you want to switch to the single `master` branch also for walking-controllers please reply, so we do not merge this PR.

Since now ``master`` and ``devel`` are aligned we can also move to the pattern release/``master``. What do you think?

traversaro:
> > @GiulioRomualdi @S-Dafarra I opened this as I saw the master/devel pattern used in walking-controllers, if you want to switch to the single `master` branch also for walking-controllers please reply, so we do not merge this PR.
> 
> Since now `master` and `devel` are aligned we can also move to the pattern release/`master`. What do you think?

That is fine for me.

GiulioRomualdi:
I also agree to keep only master.

traversaro:
Ok!

PR #1019, opened 2022-02-10
Specify that you need to also run mamba init to run some mamba commands

traversaro:
Fix problem detected by https://github.com/ami-iit/training-conda-robotics/pull/5 .

### DIFF
--- a/doc/install-mambaforge.md
+++ b/doc/install-mambaforge.md
@@ -22,7 +22,7 @@ sh ./Mambaforge-$(uname)-$(uname -m).sh -b
 This will install mambaforge in `~/mambaforge` .
 
 To use the `conda` and `mamba` commands, you need to add the `~/mambaforge/condabin` directory to the [`PATH` environment variable](https://en.wikipedia.org/wiki/PATH_(variable)). 
-You can do this persistently by modifying the `.bashrc` via the commands: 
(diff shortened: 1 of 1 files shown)


doc/install-mambaforge.md
```diff
@@ -22,14 +22,17 @@ sh ./Mambaforge-$(uname)-$(uname -m).sh -b
 This will install mambaforge in `~/mambaforge` .
 
 To use the `conda` and `mamba` commands, you need to add the `~/mambaforge/condabin` directory to the [`PATH` environment variable](https://en.wikipedia.org/wiki/PATH_(variable)). 
-You can do this persistently by modifying the `.bashrc` via the command: 
+You can do this persistently by modifying the `.bashrc` via the commands:
```
kouroshD, line 25: It was a bit misleading to me this sentence, since we need to run it on terminal once, and it will make modifications on `bashrc` automatically persistently. But this command is not necessary to be added to the `bashrc` and be called everytime.
```suggestion
You can do this persistently by modifying the `.bashrc` via writing following commands on the terminal: 
```

kouroshD: Probably should be similarly applied on other places where similar command is called.

kouroshD:
Thank you @traversaro . Looks fine to me.

PR #1023, opened 2022-02-14
Add swig to conda-forge source installation dependencies

RiccardoGrieco:
cc @traversaro

### DIFF
--- a/doc/conda-forge.md
+++ b/doc/conda-forge.md
@@ -112,7 +112,7 @@ of the robotology-superbuild.**
 Once you activated it, you can install packages in it. In particular the dependencies for the robotology-superbuild can be installed as:
 ~~~
 mamba install -c conda-forge cmake compilers make ninja pkg-config
-mamba install -c conda-forge ace asio assimp boost eigen gazebo glew glfw graphviz gsl ipopt irrlicht libjpeg-turbo libmatio libode libxml2 nlohmann_json opencv pkg-config portaudio qt sdl sdl2 sqlite tinyxml spdlog lua icu=68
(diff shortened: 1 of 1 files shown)


RiccardoGrieco:
> - Can you also add the installation to the packages installed in CI (i.e. in https://github.com/robotology/robotology-superbuild/blob/master/.github/workflows/ci.yml )

I see it installed in https://github.com/robotology/robotology-superbuild/blob/e2e907800b3a3781211842ae8362a713e93cbbd1/.github/workflows/ci.yml#L79-L82 so I think this doesn't need to be updated.

traversaro:
> > * Can you also add the installation to the packages installed in CI (i.e. in https://github.com/robotology/robotology-superbuild/blob/master/.github/workflows/ci.yml )
> 
> I see it installed in
> 
> https://github.com/robotology/robotology-superbuild/blob/e2e907800b3a3781211842ae8362a713e93cbbd1/.github/workflows/ci.yml#L79-L82
> so I think this doesn't need to be updated.

But that line mirrors the docs in https://github.com/robotology/robotology-superbuild/blob/master/doc/cmake-options.md#python . If we just want to move it to the general list as it is more user friendly, that is ok for me, but should'n we move also the other python dependencies?

RiccardoGrieco:
I opened the PR after @antonellopaolino and (back in time) also @Gianlucamilani had issues with the superbuild because `swig` was missing. 

After talking with @antonellopaolino, turns out he actually had enabled the `ROBOTOLOGY_USES_PYTHON/MATLAB` option without the related dependencies, while for @Gianlucamilani we cannot check currently the status of the options of his superbuild.

At this point I'm closing the PR without merging.

PR #1024, opened 2022-02-14
[code trigger]: changed the repository dispatch

Nicogene:
It now sends just the new released version, this is needed for an internal refactoring of `icub-tech-iit/code` actions

### DIFF
--- a/.github/workflows/icubTechIIT_code.yml
+++ b/.github/workflows/icubTechIIT_code.yml
@@ -4,7 +4,7 @@ name: Webhook trigger for code repository
 
 on:
   release:
-      types: [published, edited]
+      types: [published]
 
 jobs:
     job_id:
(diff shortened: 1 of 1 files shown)


traversaro:
> It now sends just the new released version, this is needed for an internal refactoring of `icub-tech-iit/code` actions

What happens if I edit a pre-released release? That may happen quite a lot before I actually publish a release.

Nicogene:
> > It now sends just the new released version, this is needed for an internal refactoring of `icub-tech-iit/code` actions
> 
> What happens if I edit a pre-released release? That may happen quite a lot before I actually publish a release.

It should trigger the action at every edit, and since it would be a problem on our side since we are creating tags that already exists we decided to leave just the `published` event. cc @AlexAntn @vvasco

traversaro:
Ok!

traversaro (approved):

PR #1022, opened 2022-02-13
Temporary workaround for Gazebo conda crash issue

traversaro:
See https://github.com/conda-forge/gazebo-feedstock/issues/119 for more details.

PR #1025, opened 2022-02-14
Conda: Specify explicitly requirement of qt=5.12.9=*_4

traversaro:
It is a workaround for https://github.com/conda-forge/gazebo-feedstock/issues/119 like https://github.com/robotology/robotology-superbuild/pull/1022, but I like it more because it is more explicit. Furthermore, I also adding it in the CI that I forgot in https://github.com/robotology/robotology-superbuild/pull/1022 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
         # Compilation related dependencies 
         mamba install cmake compilers make ninja pkg-config
         # Actual dependencies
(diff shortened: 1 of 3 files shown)


traversaro:
Let's wait for https://github.com/robotology/robotology-superbuild/pull/1026 to get merged before, otherwise we can't see if the conda ci gets compile as expected.

traversaro:
Thanks @kouroshD @HosameldinMohamed !

Issue #1020, opened 2022-02-11
Conda CI Failure Mid February 2022


traversaro:
Error:
~~~
2022-02-11T02:12:39.2514067Z [command]/usr/share/miniconda3/condabin/mamba create --name test
2022-02-11T02:12:39.8237286Z ##[warning]
# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/usr/share/miniconda3/lib/python3.9/site-packages/conda/exceptions.py", line 1080, in __call__
        return func(*args, **kwargs)
      File "/usr/share/miniconda3/lib/python3.9/site-packages/mamba/mamba.py", line 934, in exception_converter
        raise e
      File "/usr/share/miniconda3/lib/python3.9/site-packages/mamba/mamba.py", line 928, in exception_converter
        exit_code = _wrapped_main(*args, **kwargs)
      File "/usr/share/miniconda3/lib/python3.9/site-packages/mamba/mamba.py", line 882, in _wrapped_main
        result = do_call(args, p)
      File "/usr/share/miniconda3/lib/python3.9/site-packages/mamba/mamba.py", line 748, in do_call
        exit_code = create(args, parser)
      File "/usr/share/miniconda3/lib/python3.9/site-packages/mamba/mamba.py", line 624, in create
        return install(args, parser, "create")
      File "/usr/share/miniconda3/lib/python3.9/site-packages/mamba/mamba.py", line 577, in install
        if not transaction.prompt():
    UnboundLocalError: local variable 'transaction' referenced before assignment

`$ /usr/share/miniconda3/condabin/mamba create --name test`

  environment variables:
                 CIO_TEST=<not set>
                    CONDA=/usr/share/miniconda3
           CONDA_PKGS_DIR=/home/runner/conda_pkgs_dir
               CONDA_ROOT=/usr/share/miniconda3
           CURL_CA_BUNDLE=<not set>
      DEPLOYMENT_BASEPATH=/opt/runner
        GITHUB_EVENT_PATH=/home/runner/work/_temp/_github_workflow/event.json
              GITHUB_PATH=/home/runner/work/_temp/_runner_file_commands/add_path_f573de62-4c7d-4
                          ce0-8c28-fa0c51a3d380
                     PATH=/usr/share/miniconda3/condabin:/home/linuxbrew/.linuxbrew/bin:/home/li
[...]

PR #1026, opened 2022-02-14
Pin miniforge version to avoid mamba regression

traversaro:
The issue described in https://github.com/robotology/robotology-superbuild/issues/1020 is taking some time to be fixed upstream, so for the time being let's just use a old `Mambaforge` that is known to work.


traversaro:
The workaround seems to be working well, I think we can just merge so that we can rebase the other PRs on the top of this one. @RiccardoGrieco

RiccardoGrieco (approved):

PR #1030, opened 2022-02-14
Add generation of conda metapackages robotology-distro and robotology-distro-all for each new distro release

traversaro:
This PR adds the logic to generate the `robotology-distro` and `robotology-distro-all` conda metapackages for each new robotology-superbuild. 

This would permit user to run:
~~~
mamba install -c conda-forge -c robotology robotology-distro=2022.0 yarp icub-main idyntree
~~~

And the solver will automatically compute the correct version of `yarp`, `icub-main` and `idyntree` to install. Note that `robotology-distro` will just constrain the version of packages that are installed by the user, so installing just that package will not install anything else. On the other hand, this PR also adds the `robotology-distro-all` metapackage, that will install all the packages contained in the robotology-superbuild corresponding to a given release.

No documentation is currently added as I prefer to actually have the first metapackages generated and present on the `robotology` channel before document this. 

In details, this PR:
* Also generate conda packages on releases and not only on weekly jobs on master branch, to make sure that the versions references by the metapackages actually exists
* Read and write `CONDA_BUILD_NUMBER` always on `master` branch, even when the conda packages are generated on a branch (that is tipically is a `release/YYYY.MM`
* On releases, also generate the robotology-distro and robotology-distro-all metapackages . This is done by appropriate modifications to the conda packages generation logic, that is described in https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-recipe-generation.md#internals

Fix https://github.com/robotology/robotology-superbuild/issues/727 .

### DIFF
--- a/.github/workflows/generate-conda-packages.yaml
+++ b/.github/workflows/generate-conda-packages.yaml
@@ -114,9 +114,6 @@ jobs:
           if: contains(matrix.os, 'macos') || contains(matrix.os, 'ubuntu')
           shell: bash -l {0}
           run: |
-            echo "==============="
-            echo ${{needs.get-conda-build-number.outputs.conda_build_number}}
-            echo "==============="
             mkdir build
             cd build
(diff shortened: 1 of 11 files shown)


traversaro:
fyi @GiulioRomualdi

Nicogene (approved):

Nicogene:
I am not a conda expert, but I like very much this addition 👍🏻

traversaro:
Thanks @Nicogene !

Issue #1029, opened 2022-02-14
Ubuntu Jammy CI Failure Mid February 2022


traversaro:
Failure in `find_package(gazebo)`:
~~~
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/gazebo/gazebo-config.cmake:239 (find_package):
  By not providing "Findignition-transport8.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "ignition-transport8", but CMake did not find one.

  Could not find a package configuration file provided by
  "ignition-transport8" with any of the following names:

    ignition-transport8Config.cmake
    ignition-transport8-config.cmake

  Add the installation prefix of "ignition-transport8" to CMAKE_PREFIX_PATH
  or set "ignition-transport8_DIR" to a directory containing one of the above
  files.  If "ignition-transport8" provides a separate development package or
  SDK, be sure it has been installed.
Call Stack (most recent call first):
  cmake/BuildGazeboYARPPlugins.cmake:31 (find_package)
  build/install/share/YCM/modules/FindOrBuildPackage.cmake:204 (include)
  cmake/RobotologySuperbuildLogic.cmake:19 (find_or_build_package)
  CMakeLists.txt:63 (include)


-- Configuring incomplete, errors occurred!
See also "/__w/robotology-superbuild/robotology-superbuild/build/CMakeFiles/CMakeOutput.log".
~~~

traversaro:
ubuntu:jammy job removed in https://github.com/robotology/robotology-superbuild/pull/1036 .

PR #1031, opened 2022-02-14
ci: Ensure that gazebo is not searched in ubuntu:jammy

traversaro:
Workaround for https://github.com/robotology/robotology-superbuild/issues/1029 . As we need to set `ROBOTOLOGY_USES_GAZEBO` is set to `OFF` since the beginning, I had to remove the use of `${GITHUB_WORKSPACE}/.ci/initial-cache.gh.cmake` and listed all the options manually.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,7 +215,7 @@ jobs:
       run: |
         mkdir -p build
         cd build
-        cmake -C ${GITHUB_WORKSPACE}/.ci/initial-cache.gh.cmake  -G"${{ matrix.cmake_generator }}"  -DROBOTOLOGY_USES_GAZEBO:BOOL=OFF -DNON_INTERACTIVE_BUILD:BOOL=TRUE -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}  ${{ matrix.project_tags_cmake_options }}  ..
+        cmake -DROBOTOLOGY_USES_GAZEBO:BOOL=OFF -C ${GITHUB_WORKSPACE}/.ci/initial-cache.gh.cmake  -G"${{ matrix.cmake_generator }}"  -DROBOTOLOGY_USES_GAZEBO:BOOL=OFF -DNON_INTERACTIVE_BUILD:BOOL=TRUE -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}  ${{ matrix.project_tags_cmake_options }}  ..
         # Octave is disabled as a workaround for https://github.com/robotology/robotology-superbuild/issues/384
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -211,17 +211,15 @@ jobs:
...
         apt-get -y upgrade
 
-
-
     - name: Configure [Docker]
       run: |
         mkdir -p build
         cd build
-        cmake -C ${GITHUB_WORKSPACE}/.ci/initial-cache.gh.cmake  -G"${{ matrix.cmake_generator }}" -DNON_INTERACTIVE_BUILD:BOOL=TRUE -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}  ${{ matrix.project_tags_cmake_options }}  ..
+        cmake -DROBOTOLOGY_USES_GAZEBO:BOOL=OFF -C ${GITHUB_WORKSPACE}/.ci/initial-cache.gh.cmake  -G"${{ matrix.cmake_generator }}"  -DROBOTOLOGY_USES_GAZEBO:BOOL=OFF -DNON_INTERACTIVE_BUILD:BOOL=TRUE -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}  ${{ matrix.project_tags_cmake_options }}  ..
```
RiccardoGrieco, line 218:
```suggestion
        cmake  -C ${GITHUB_WORKSPACE}/.ci/initial-cache.gh.cmake  -G"${{ matrix.cmake_generator }}" -DROBOTOLOGY_USES_GAZEBO:BOOL=OFF -DNON_INTERACTIVE_BUILD:BOOL=TRUE -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}  ${{ matrix.project_tags_cmake_options }}  ..
```

Is there a reason why `-DROBOTOLOGY_USES_GAZEBO:BOOL=OFF` is repeated?

traversaro: I tried to set it both before and after and the `-C ${GITHUB_WORKSPACE}/.ci/initial-cache.gh.cmake ` parameter as the initial cache there instead sets ROBOTOLOGY_USES_GAZEBO to `ON`, but apparently this is not possible. Let's find another solution.

traversaro:
> Argh, icub-models is still failing as it is silently looking for Gazebo https://github.com/robotology/icub-models/blob/c73aa1d7e42e571c0b5b9f12043a07f9a1377e3f/CMakeLists.txt#L40 . At this point, better just disable jammy and re-enable it at some point in the future (hopefully before its release).

See https://github.com/robotology/robotology-superbuild/pull/1036 .

PR #1036, opened 2022-02-15
Disable CI tests on ubuntu:jammy

traversaro:
Workaround for https://github.com/robotology/robotology-superbuild/issues/1029, as the problem is in Ubuntu Jammy and not on our side.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,7 +165,8 @@ jobs:
         docker_image:
           - "ubuntu:bionic"
           - "ubuntu:focal"
-          - "ubuntu:jammy"
+          # Disabled as workaround for https://github.com/robotology/robotology-superbuild/issues/1029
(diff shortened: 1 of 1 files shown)


RiccardoGrieco:
If https://github.com/robotology/icub-models/pull/129 is merged, we may not have the same issue experienced in #1031 when disabling `ROBOTOLOGY_USES_GAZEBO`.

We could check this before relying on this workaround.

traversaro:
> If [robotology/icub-models#129](https://github.com/robotology/icub-models/pull/129) is merged, we may not have the same issue experienced in #1031 when disabling `ROBOTOLOGY_USES_GAZEBO`.
> 
> We could check this before relying on this workaround.

The problem is that that PR is fixing the issue in `devel`, and for various reason in that repo we only merge devel in master at releases. So that PR is not a short term solution.

RiccardoGrieco (approved):

Issue #1046, opened 2022-02-25
2022.02 release CI dry run


traversaro:
I did a 2022.02 release dry run in:
* C++ CI: https://github.com/traversaro/robotology-superbuild/actions/runs/1900139386
* Generate Conda Packages CI: https://github.com/traversaro/robotology-superbuild/actions/runs/1900139387

To check if the CI (that is quite more complex since https://github.com/robotology/robotology-superbuild/pull/1030) is working fine.
The release dry run will not interfere with the regular releases as it is done in a fork.

traversaro:
These are the problems:

## P1

Conda packages generation on macOS, as described in https://github.com/robotology/robotology-superbuild/issues/1041 .

## P2 

Conda **meta**packaages generation on Windows and Linux:
~~~
2022-02-25T22:18:40.6574848Z conda-forge/linux-64     Using cache
2022-02-25T22:18:40.6599134Z conda-forge/noarch       Using cache
2022-02-25T22:18:41.0667520Z Reloading output folder: /usr/share/miniconda/envs/test/conda-bld
2022-02-25T22:18:41.1710391Z Reloading output folder: /usr/share/miniconda/envs/test/conda-bld
2022-02-25T22:18:41.3220950Z BUILD START: ['robotology-distro-all-2022.02.0-ha770c72_43.tar.bz2']
2022-02-25T22:18:41.3221854Z Reloading output folder: /usr/share/miniconda/envs/test/conda-bld
2022-02-25T22:18:41.4236448Z Reloading output folder: /usr/share/miniconda/envs/test/conda-bld
2022-02-25T22:18:42.4082529Z Mamba failed to solve:
2022-02-25T22:18:42.4083178Z  - blocktest 2.3.4.*
2022-02-25T22:18:42.4083544Z  - yarp-matlab-bindings 3.6.0.*
2022-02-25T22:18:42.4083926Z  - whole-body-estimators 0.6.1.*
2022-02-25T22:18:42.4084304Z  - gazebo-yarp-plugins 4.1.2.*
2022-02-25T22:18:42.4084617Z  - wb-toolbox 5.5.0.*
2022-02-25T22:18:42.4084979Z  - unicycle-footstep-planner 0.4.1.*
2022-02-25T22:18:42.4085340Z  - yarp-telemetry 0.4.0.*
2022-02-25T22:18:42.4085664Z  - yarp-device-xsensmt 0.1.1.*
[...]

PR #1049, opened 2022-02-26
macOS conda package generation: Pin gazebo and opencv and do not generate metapackages

traversaro:
The pin was added in https://github.com/robotology/robotology-superbuild/pull/1025 to avoid any problem related to https://github.com/conda-forge/gazebo-feedstock/issues/119 . Anyhow, this is creating problems in macOS, so add two further pins and disable the metapackage generation.


traversaro:
Test `generate-conda-packages` build: https://github.com/robotology/robotology-superbuild/runs/5345237905 .

traversaro:
This is actually not working, something different is being tested as described in https://github.com/robotology/robotology-superbuild/issues/1046#issuecomment-1053576161 .

Issue #1057, opened 2022-03-06
Compilation issues when installing the superbuild on windows from sources with conda


gabrielenava:
I followed the installation instructions in https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-forge.md#source-installation to install the superbuild on Windows 10 with options `ROBOTOLOGY_USES_MATLAB` and `ROBOTOLOGY_USES_DYNAMICS` both set to `ON`. I am using "Visual Studio 17 2022" as compiler.

I noted that after I compile the superbuild, several folders are created inside the `src` folder, but they are actually empty (no git clone performed).   The compilation apparently terminates with no final error messages.

The empty repos are: `whole-body-controllers`, `WBToolbox`, `idyntree-yarp-tools`, `matlab-whole-body-simulator`, `whole-body-estimators`.

 I have then checked the full output of the terminal from the moment I run `cmake`, and I spotted two error messages when compiling YARP:

```cmake

 yarpcar.vcxproj -> C:\Users\gnava\Documents\GitHub\robotology-superbuild\build\src\YARP\bin\Debug\yarpcard.dll
    main.cpp
main.obj : error LNK2019: unresolved external symbol "void __cdecl cv::cvtColor(class cv::debug_build_guard::_InputArra
y const &,class cv::debug_build_guard::_OutputArray const &,int,int)" (?cvtColor@cv@@YAXAEBV_InputArray@debug_build_gua
rd@1@AEBV_OutputArray@31@HH@Z) referenced in function "class cv::Mat __cdecl yarp::cv::toCvMat<unsigned char>(class yar
p::sig::ImageOf<unsigned char> &)" (??$toCvMat@E@cv@yarp@@YA?AVMat@0@AEAV?$ImageOf@E@sig@1@@Z) [C:\Users\gnava\Document
s\GitHub\robotology-superbuild\build\src\YARP\src\yarpdatadumper\yarpdatadumper.vcxproj] [C:\Users\gnava\Documents\GitH
ub\robotology-superbuild\build\YARP.vcxproj]
C:\Users\gnava\Documents\GitHub\robotology-superbuild\build\src\YARP\bin\Debug\yarpdatadumper.exe : fatal error LNK1120
: 1 unresolved externals [C:\Users\gnava\Documents\GitHub\robotology-superbuild\build\src\YARP\src\yarpdatadumper\yarpd
[...]

PR #1059, opened 2022-03-09
Update conda-forge.md

gabrielenava:
Updated documentation as per https://github.com/robotology/robotology-superbuild/issues/1057#issuecomment-1062705144

### DIFF
--- a/doc/conda-forge.md
+++ b/doc/conda-forge.md
@@ -155,6 +155,8 @@ cmake --build . --config Release
 
 **IMPORTANT: conda-forge does not provide Debug version of its libraries, so in Windows you can't compile in Debug mode if you are using conda-forge.**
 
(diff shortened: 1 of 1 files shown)


traversaro:
I am inclined to not include such a particular statement in the documentation. New users that will install CMake will almost certainly install the latest version of CMake, so they will not get this problem. On the other hand, adding new details on the documentation is making the documentation even less readable, so I think probably we can avoid adding this?

gabrielenava:
>I am inclined to not include such a particular statement in the documentation. New users that will install CMake will almost certainly install the latest version of CMake, so they will not get this problem. On the other hand, adding new details on the documentation is making the documentation even less readable, so I think probably we can avoid adding this?

agreed, closing without merging!

PR #1070, opened 2022-03-24
BuildYARP: Set ENABLE_yarpcar_mjpeg to ON on Apple if we are using Unstable branches

traversaro:
As the unstable branches contain the fix  https://github.com/robotology/yarp/pull/2810 .

FYI @S-Dafarra

### DIFF
--- a/cmake/BuildYARP.cmake
+++ b/cmake/BuildYARP.cmake
@@ -31,7 +31,8 @@ else()
 endif()
 
 # Workaround for https://github.com/robotology/yarp/issues/2353
-if(APPLE AND NOT ${ROBOTOLOGY_CONFIGURING_UNDER_CONDA})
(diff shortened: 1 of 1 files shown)


traversaro:
Intersting, conda@macOS fails exactly with the same error. Probably on conda both jpeg and libjpeg-turbo are installed, but while on Linux and Windows the jpeg-turbo header is used by the preprocessor, on macOS the jpeg one is found hence the problem that was fixed in https://github.com/robotology/yarp/pull/2810 but still needs to be released. Probably we can change this to enable the option if Unstable branches are used.

traversaro:
Friendly ping @RiccardoGrieco @nunoguedelha

RiccardoGrieco (approved):

PR #1078, opened 2022-03-30
 Migrate YCMBootstrap to use FetchContent module 

traversaro:
Migrate to use YCMBootstrapFetch introduced in https://github.com/robotology/ycm/pull/403 and released as part of YCM 0.14.0 .

As this is the last step of https://github.com/robotology/robotology-superbuild/issues/1091, this also fixes https://github.com/robotology/robotology-superbuild/issues/1091 .


traversaro:
The test was successful.

Nicogene (approved):

Issue #1008, opened 2022-01-31
Unable to find `soxr.h` (from conda-forge dependencies installation on Manjaro)


HosameldinMohamed:
Very similar to #1002. And with the same setup procedure.

Probably `conda-forge` dependencies are missing `soxr`.

```
/home/hosameldin/dev/robotology-superbuild/src/YARP/src/libYARP_sig/src/yarp/sig/SoundFilters.cpp:21:10: fatal error: soxr.h: No such file or directory
   21 | #include <soxr.h>
```

But in this case, the header `soxr.h` is available Conda in the channel [`HCC`](https://anaconda.org/hcc/soxr). If this package is fine maybe we add it to the [documentation](https://github.com/robotology/robotology-superbuild/blob/master/doc/conda-forge.md#source-installation).

HosameldinMohamed:
CC @traversaro

traversaro:
Thanks @HosameldinMohamed for reporting the issue! Not that probably this is due to https://github.com/robotology/yarp/pull/2665, and soxr is not installed even in apt/homebrew/vcpkg .  For what regards the conda package, I would avoid to depend on packages that are not part of `conda-forge` (or stricly related channels like `robostack` or `robotology`). The reason is that packages in other channels may not be compatible with the one from conda-forge, for example because they duplicate some dependency, or because they do not support some platforms supported by us and conda-forge. For example, the HCC channel does not support Windows and macOS. As soxr is quite a simple CMake-based project, it should not be difficult to add it on conda-forge.

HosameldinMohamed:
> As soxr is quite a simple CMake-based project, it should not be difficult to add it on conda-forge, taking the example from https://github.com/conda-forge/staged-recipes/pull/15408/files or similar simple CMake-package recipes.

Cool! I'll give it a try!

traversaro:
> > As soxr is quite a simple CMake-based project, it should not be difficult to add it on conda-forge, taking the example from https://github.com/conda-forge/staged-recipes/pull/15408/files or similar simple CMake-package recipes.
> 
> Cool! I'll give it a try!

Great, let me know if you need an hand.

traversaro:
Cool!

> The one thing I don't understand is how to add a command to extract the package?

[...]

PR #1087, opened 2022-04-08
Update Conda documentation - add Soxr dependency

HosameldinMohamed:
This should fix https://github.com/robotology/robotology-superbuild/issues/1008

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
         # Compilation related dependencies 
         mamba install cmake compilers make ninja pkg-config
         # Actual dependencies
-        mamba install ace asio assimp boost eigen gazebo glew glfw graphviz gsl ipopt irrlicht libjpeg-turbo libmatio libode libxml2 nlohmann_json opencv pkg-config portaudio qt=5.12.9=*_4 sdl sdl2 sqlite tinyxml spdlog lua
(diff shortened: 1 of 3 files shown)


traversaro:
Can you add the installation also in CI, see https://github.com/HosameldinMohamed/robotology-superbuild/blob/updateDocSoxr/.github/workflows/ci.yml#L80 ? Unfortunatly we still have this info duplicated.

HosameldinMohamed:
> Can you add the installation also in CI, see https://github.com/HosameldinMohamed/robotology-superbuild/blob/updateDocSoxr/.github/workflows/ci.yml#L80 ? Unfortunatly we still have this info duplicated.

I had missed that. done in https://github.com/robotology/robotology-superbuild/pull/1087/commits/b924f2215fd9c1a65e36bd9fe75e373d5021e079

traversaro:
Please ignore the Unstable CI failures. They are regression of https://github.com/robotology/ycm/pull/401 that will be fixed by https://github.com/robotology/icub-main/pull/787 .

traversaro:
For consistency I would also update the apt dependencies adding `libsoxr-dev` in https://github.com/robotology/robotology-superbuild/blob/master/apt.txt (fortunatly this list is used both by docs and CI, so in this case you just need to update a single file).

HosameldinMohamed:
> For consistency I would also update the apt dependencies adding `libsoxr-dev` in https://github.com/robotology/robotology-superbuild/blob/master/apt.txt (fortunatly this list is used both by docs and CI, so in this case you just need to update a single file).

done in https://github.com/robotology/robotology-superbuild/pull/1087/commits/f8001667ad1fd0aca317aca20f4717be306860b6

traversaro:
Thanks @HosameldinMohamed !

PR #1102, opened 2022-04-27
Check if SWIG is installed when enabling ROBOTOLOGY_USES_PYTHON

RiccardoGrieco:
This PR to check if `SWIG` is installed when the `ROBOTOLOGY_USES_PYTHON` option is enabled.

In this way we can prevent failing builds at the superbuild config steps.


RiccardoGrieco:
cc @traversaro

PR #556, opened 2020-12-13
Clone all the repos in the src directory instead of external and robotology

traversaro:
This was discussed in https://github.com/robotology/robotology-superbuild/issues/333 . 
As this is kind an important change, it will be announced on https://github.com/robotology/QA before the PR is merged.

PR #1117, opened 2022-05-13
Fix path for icub-tests context

gsisinna:
This fix the regression from https://github.com/robotology/robotology-superbuild/pull/556. We tested it with:

`yarp resource --context icubEth --from joint_limits_face.ini`

Now `$YARP_DATA_DIRS` env variable points to the correct folder for icub-tests contexts.

cc @valegagge


traversaro (approved):

traversaro:
Thanks @gsisinna ! 


Just for the sake of completeness, it useful to mentione that ideally we should not mention the source directory at all in the `setup.sh`, to avoid this kind of problems. See https://github.com/robotology/icub-tests/issues/57#issuecomment-877278389 .

traversaro:
I just realized that there are several other points that needs the same fix, see:
* https://github.com/gsisinna/robotology-superbuild/blob/19672f0f2c3da27e2e3768c68c322ebcdccc8b15/cmake/template/setup.bat.in#L30
* https://github.com/gsisinna/robotology-superbuild/blob/master/cmake/template/addPathsToUserEnvVariables.ps1.in#L63
* https://github.com/gsisinna/robotology-superbuild/blob/master/cmake/template/removePathsFromUserEnvVariables.ps1.in#L66

traversaro (changes requested):
Do you think you can update https://github.com/robotology/robotology-superbuild/pull/1117#issuecomment-1126076312 ?

traversaro (approved):
Thanks a lot!

PR #1129, opened 2022-05-26
Switch PROJECT_TAGS Stable to use YARP 3.7

traversaro:
YARP 3.7.0 was released recently: https://github.com/robotology/community/discussions/579 .


Nicogene (approved):

traversaro:
The branch is not merging as `walking-teleoperation` master branch is not compiling with YARP 3.7.0 . However, I think it is better to merge so we can fix https://github.com/robotology/robotology-superbuild/issues/1133 for all components different from teleoperation.

PR #1130, opened 2022-05-26
Update latest.releases.yaml to `icub-firmware-models@v1.25.0`

pattacini:
As per the title.

### DIFF
--- a/releases/latest.releases.yaml
+++ b/releases/latest.releases.yaml
@@ -142,7 +142,7 @@ repositories:
   icub-firmware-models:
     type: git
     url: https://github.com/robotology/icub-firmware-models.git
-    version: v1.20.0
+    version: v1.25.0
   yarp-device-xsensmt:
     type: git
     url: https://github.com/robotology/yarp-device-xsensmt.git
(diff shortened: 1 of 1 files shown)


Nicogene (approved):
LGTM but maybe we have to await https://github.com/robotology/robotology-superbuild/pull/1129?

Nicogene:
I manually updated also `icub-firmware` and `icub-firmware-build` since they are not automatically updated by the GA

pattacini:
> I manually updated also `icub-firmware` and `icub-firmware-build` since they are not automatically updated by the GA

Ah didn't know about that, thus I updated manually the `icub-firmware-shared` too.
Hope this won't cause trouble.

Nicogene:
> > I manually updated also `icub-firmware` and `icub-firmware-build` since they are not automatically updated by the GA
> 
> Ah didn't know about that, thus I updated manually the `icub-firmware-shared` too. Hope this won't cause trouble.

I should be not, the version is automatically updated by those repo that compiles, for example icub-firmware does not then we have to remember to update it manually

traversaro:
Thanks, merging as all failures are due to:
* https://github.com/robotology/robotology-superbuild/issues/1133
* https://github.com/robotology/robotology-superbuild/issues/1134

Not related to this PR.

PR #1145, opened 2022-06-01
Switch MATLAB one-line install to mamba

traversaro:
Let's check if https://github.com/robotology/robotology-superbuild/pull/1144 is working fine or not even if we switch to mamba.

### DIFF
--- a/.github/workflows/matlab-one-line-install-test.yml
+++ b/.github/workflows/matlab-one-line-install-test.yml
@@ -49,6 +49,12 @@ jobs:
             cd scripts
             install_robotology_packages()
             
+      # workaround for https://github.com/robotology/robotology-superbuild/issues/64
+      # and https://github.com/robotology/idyntree/issues/995
+      - name: Do not use MATLAB's stdc++ to avoid incompatibilities with other libraries
+        if: contains(matrix.os, 'ubuntu')
+        run:
+          echo "LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libstdc++.so.6" >> $GITHUB_ENV
      
       - name: Run MATLAB commands
         uses: matlab-actions/run-command@v1
--- a/.github/workflows/matlab-one-line-install-test.yml
+++ b/.github/workflows/matlab-one-line-install-test.yml
@@ -54,5 +54,6 @@ jobs:
         uses: matlab-actions/run-command@v1
         with:
           command: |
+            robotology_setup
             pos = iDynTree.Position()
             vec = yarp.Vector()
--- a/.github/workflows/matlab-one-line-install-test.yml
+++ b/.github/workflows/matlab-one-line-install-test.yml
@@ -54,6 +54,7 @@ jobs:
         uses: matlab-actions/run-command@v1
         with:
           command: |
+            cd scripts
             robotology_setup
             pos = iDynTree.Position()
             vec = yarp.Vector()
--- a/.github/workflows/matlab-one-line-install-test.yml
+++ b/.github/workflows/matlab-one-line-install-test.yml
@@ -1,13 +1,18 @@
 name: MATLAB tests
 
 on:
+  workflow_dispatch:
   push:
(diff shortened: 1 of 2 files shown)


traversaro:
@FabioBergonti it is still working fine (see https://github.com/robotology/robotology-superbuild/runs/6690394514). Do you have any idea on what could be different w.r.t. to your case?

FabioBergonti:
> @FabioBergonti it is still working fine (see https://github.com/robotology/robotology-superbuild/runs/6690394514). Do you have any idea on what could be different w.r.t. to your case?

No!
To debug I added a [step](`https://github.com/ami-iit/paper_bergonti_2022_tro_kinematics-control-morphingcovers/runs/6752227347?check_suite_focus=true#step:4:336`) in my CI where I install robotology via matlab

``` yaml
      - name: Test Robotology
        uses: matlab-actions/run-command@v1
        with:
          command: |
            websave('install_robotology_packages.m', 'https://raw.githubusercontent.com/robotology/robotology-superbuild/onelineinstallswitchtomamba/scripts/install_robotology_packages.m')
            install_robotology_packages
            system("cat  /home/runner/work/paper_bergonti_2022_tro_kinematics-control-morphingcovers/paper_bergonti_2022_tro_kinematics-control-morphingcovers/robotology-matlab/condabin/mamba")
            robotology_setup
            pos = iDynTree.Position()
            vec = yarp.Vector()
```

Unfortunately I have the same problem (see [action result](
https://github.com/ami-iit/paper_bergonti_2022_tro_kinematics-control-morphingcovers/runs/6752401459?check_suite_focus=true#step:4:359)).
``` matlab
  Traceback (most recent call last):
    File "/home/runner/work/paper_bergonti_2022_tro_kinematics-control-morphingcovers/paper_bergonti_2022_tro_kinematics-control-morphingcovers/robotology-matlab/condabin/mamba", line 7, in <module>
      from mamba.mamba import main
```

output `cat  .../robotology-matlab/condabin/mamba`:
```
  #!/usr/bin/env python
  
  # -*- coding: utf-8 -*-
  import re
  import sys
  
  from mamba.mamba import main
  
  if __name__ == '__main__':
      sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
      sys.exit(main())
```

traversaro:
Comparing the logs the only difference seems to be the lenght of the prefix, 164 in the case of @FabioBergonti and 88 in the case of this PR.

FabioBergonti:
> Bingo! The problem is reproduced by using a 167 character long prefix.

I confirm that reducing the length of the prefix solved the problem (see [action results](https://github.com/ami-iit/paper_bergonti_2022_tro_kinematics-control-morphingcovers/runs/6753335823?check_suite_focus=true))

traversaro:
The same problem occurs indipently of wheter conda or mamba is used. As micromamba (basically the pure C++ reimplementation of conda is used, as opposed to mamba that is a mix between some python conda code with some C++ code) is used in mambaforge, I suspect the problem is in micromamba. I will try to open an issue there as the devs may have more insight.

traversaro:
@FabioBergonti found that this is actually intentional, see https://github.com/mamba-org/mamba/issues/1731 . To solve this problem we could:
* In the short term, print an error if the prefix used is too long
* In the longer term, switch to micromamba for this use case, so we avoid completly the need to call mamba or conda python scripts with all the related problems of shebang length
* An alternative solution at the micromamba level could be to have two scripts, one with a short shebang that just invoke the python with the full path with the correct script. However, I am not sure how much effort would be something like that.

traversaro:
Superseded by https://github.com/robotology/robotology-superbuild/pull/1154 .

> In the longer term, switch to micromamba for this use case, so we avoid completly the need to call mamba or conda python scripts with all the related problems of shebang length

Tracked in https://github.com/robotology/robotology-superbuild/issues/1157 .

PR #1148, opened 2022-06-06
install_robotology_packages: Throw an error if PREFIX is too long

FabioBergonti:
In https://github.com/robotology/robotology-superbuild/pull/1145 we discovered that a long PREFIX may cause the usage of an incorrect  `shebang`.
This PR adds an error in case the PREFIX is too long


traversaro:
Thanks @FabioBergonti !

Issue #1139, opened 2022-05-30
CI Failure of Unstable PROJECT_TAGS of Start of June 2022  of ROBOTOLOGY_ENABLE_EVENT_DRIVEN option


traversaro:
I guess the deprecation cleanup of https://github.com/robotology/yarp/pull/2846 will trigger more unstable failures. The first one I saw is in https://github.com/robotology/robotology-superbuild/runs/6657443103?check_suite_focus=true, see:
~~~
2022-05-30T15:27:03.5628961Z FAILED: lib/CMakeFiles/event-driven.dir/src/vDraw_ISO.cpp.o 
2022-05-30T15:27:03.5630450Z /usr/bin/c++ -DCLOCK_PERIOD=80 -DCODEC_304x240_24 -DTIMER_BITS=30 -Devent_driven_EXPORTS -I/__w/robotology-superbuild/robotology-superbuild/src/event-driven/lib/include -isystem /__w/robotology-superbuild/robotology-superbuild/build/install/include -isystem /usr/include/opencv -g -fPIC -Wall -fno-strict-aliasing -std=c++17 -MD -MT lib/CMakeFiles/event-driven.dir/src/vDraw_ISO.cpp.o -MF lib/CMakeFiles/event-driven.dir/src/vDraw_ISO.cpp.o.d -o lib/CMakeFiles/event-driven.dir/src/vDraw_ISO.cpp.o -c /__w/robotology-superbuild/robotology-superbuild/src/event-driven/lib/src/vDraw_ISO.cpp
2022-05-30T15:27:03.5631683Z In file included from /__w/robotology-superbuild/robotology-superbuild/src/event-driven/lib/include/event-driven/all.h:5,
2022-05-30T15:27:03.5632332Z                  from /__w/robotology-superbuild/robotology-superbuild/src/event-driven/lib/include/event-driven/vDraw.h:25,
2022-05-30T15:27:03.5632937Z                  from /__w/robotology-superbuild/robotology-superbuild/src/event-driven/lib/src/vDraw_ISO.cpp:22:
2022-05-30T15:27:03.5633729Z /__w/robotology-superbuild/robotology-superbuild/src/event-driven/lib/include/event-driven/vCollectSend.h:31:1: error: expected class-name before '{' token
2022-05-30T15:27:03.5634117Z  {
2022-05-30T15:27:03.5634283Z  ^
[...]

PR #1160, opened 2022-06-10
Remove use of walking-teleoperation devel branch

traversaro:
As per https://github.com/robotology/walking-teleoperation/issues/101#issuecomment-1152082481, the walking-teleoperation does use anymore the `devel`  branch. 

See also https://github.com/robotology/robotology-superbuild/issues/1139#issuecomment-1152028625 .


traversaro:
Discussed in person.

Issue #1159, opened 2022-06-10
 one line matlab installer on Ubuntu 20.04 does not work out of the  box


traversaro:
After 20 hours it was merged, the test of one line matlab installer failed.

Last Successful Build: https://github.com/robotology/robotology-superbuild/runs/6810705069
First CI Fail: https://github.com/robotology/robotology-superbuild/actions/runs/2473119223

traversaro:
The error is:
~~~
  {Error using yarp.Vector
  Invalid MEX-file
  '/home/runner/work/robotology-superbuild/robotology-superbuild/scripts/robotology-matlab/mex/yarpMEX.mexa64':
  /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.30' not found
  (required by
  /home/runner/work/robotology-superbuild/robotology-superbuild/scripts/robotology-matlab/mex/../lib/libYARP_os.so.3)
  
  Error in command_751a7662_794c_4a02_ae5c_0042c0fd82b9 (line 4)
  vec = yarp.Vector()
  } 
~~~

Difference between before and after:
~~~
--- <unnamed>
+++ <unnamed>
@@ -1,28 +1,30 @@
-    Package                              Version  Build                Channel                   Size
+   Package                              Version  Build                Channel                   Size
   ─────────────────────────────────────────────────────────────────────────────────────────────────────
     Install:
   ─────────────────────────────────────────────────────────────────────────────────────────────────────
   
-    + ace                                  7.0.6  h9c3ff4c_0           conda-forge/linux-64       8MB
+    + ace                                  7.0.7  h27087fc_0           conda-forge/linux-64       7MB
     + alsa-lib                           1.2.6.1  h7f98852_0           conda-forge/linux-64     592kB
     + ampl-mp                              3.1.0  h2cc385e_1006        conda-forge/linux-64       1MB
     + aom                                  3.3.0  h27087fc_1           conda-forge/linux-64       3MB
-    + assimp                               5.2.3  h4c92fa2_1           conda-forge/linux-64       4MB
+    + assimp                               5.2.4  h4c92fa2_0           conda-forge/linux-64       4MB
     + asttokens                            2.0.5  pyhd8ed1ab_0         conda-forge/noarch        22kB
+    + attr                                 2.5.1  h166bdaf_0           conda-forge/linux-64      71kB
     + backcall                             0.2.0  pyh9f0ad1d_0         conda-forge/noarch        14kB
[...]

PR #1162, opened 2022-06-10
Switch to ubuntu-22.04 for workaround of matlab one line install job

traversaro:
Workaround for https://github.com/robotology/robotology-superbuild/issues/1159 .


traversaro:
Great, it is working fine as Ubuntu 22.04 has a new enough std C++ library. Let's merge to fix CI.

traversaro:
> Great, it is working fine as Ubuntu 22.04 has a new enough std C++ library. Let's merge to fix CI.

Explanation note: at the moment `ubuntu-latest` is not 22.04, but 20.04 .

Issue #377, opened 2020-04-09
YARP does not build on GitHub Actions on macOS if YARP_USE_SYSTEM_SQLite is ON


traversaro:
Both if `sqlite` is installed via homebrew and if it is not installed, see: 
* https://github.com/robotology/robotology-superbuild/issues/375
* https://github.com/robotology/robotology-superbuild/pull/376

lrapetti:
I have updated `sqlite` via `homebrew` from `3.30.1` to the latest version (`3.31.1`), but still I was able to compile `YARP` with `YARP_USE_SYSTEM_SQLite ON`.

traversaro:
Thanks @lrapetti !  Then probably there is something strange going on with the macOS GitHub Actions image.

PR #1171, opened 2022-06-16
apt.txt: add libsqlite3-dev

Nicogene:
It is a YARP dependency that was missing from `apt.txt`

### DIFF
--- a/apt.txt
+++ b/apt.txt
@@ -60,3 +60,4 @@ libglfw3-dev
 libglew-dev
 libglm-dev
 libfreetype6-dev
+libsqlite3-dev


gsisinna:
I think this is the same problem I found [here](https://github.com/icub-tech-iit/proto/issues/1544#issuecomment-1156663338)

traversaro (approved):

traversaro:
Thanks @Nicogene ! We may also want to revisit https://github.com/robotology/robotology-superbuild/issues/377 .

traversaro:
> I think this is the same problem I found here:
> 
> > During the robotology-superbuild installation we found several compilation errors that were solved by manually installing the correct version of YCM (0.14) and a library for C (sqlite3, see [solution](https://stackoverflow.com/questions/28969543/fatal-error-sqlite3-h-no-such-file-or-directory) used).

Here where? The YCM problems seems unrelated.

traversaro:
Ahh, cool, thanks!

traversaro:
Cherry-picked in `releases/2022.05` in https://github.com/robotology/robotology-superbuild/commit/36e0ed5faa0553b8b2c9c6061a2c689416e31921 .

PR #1180, opened 2022-06-21
Update 2022.05.2.yaml w/ `icub-main@v1.16.0`

pattacini:
As per the title.

PR #1182, opened 2022-06-21
2022.05.2 : Bump `yarp@v3.7.1`, `icub-firmware-shared@v1.25.1`

traversaro:
This should fix the failure reported in https://github.com/robotology/robotology-superbuild/pull/1180#issuecomment-1162274510 .

### DIFF
--- a/releases/2022.05.2.yaml
+++ b/releases/2022.05.2.yaml
@@ -30,7 +30,7 @@ repositories:
   YARP:
     type: git
     url: https://github.com/robotology/yarp.git
-    version: v3.7.0
+    version: v3.7.1
   ICUB:
     type: git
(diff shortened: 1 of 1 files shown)


traversaro:
Probably we need to bump icub-firmware-shared too: 
~~~
2022-06-21T20:34:11.5091687Z CMake Error at conf/iCubFindDependencies.cmake:76 (message):
2022-06-21T20:34:11.5092098Z   An old version of icub-firmware-shared has been detected, but at least
2022-06-21T20:34:11.5092353Z   1.25.1 is required
2022-06-21T20:34:11.5092551Z Call Stack (most recent call first):
2022-06-21T20:34:11.5092777Z   CMakeLists.txt:57 (include)
~~~

PR #1184, opened 2022-06-22
conda: Add eigen as dependency of matio-cpp

traversaro:
Otherwise matio-cpp will be compiled without Eigen support, see https://github.com/ami-iit/matio-cpp/blob/ed9b196360b8083c3eccd3348afa7969d3a4f1fa/CMakeLists.txt#L94 .


Nicogene (approved):

traversaro:
Backported to `releases/2022.05` https://github.com/robotology/robotology-superbuild/commit/683c3489e7670898ce4854d9c2a3347f4e8a0727 .

PR #1193, opened 2022-06-28
Update forcetorque-yarp-devices name

RiccardoGrieco:
Update the name and link of the former `forcetorque-yarp-devices` to `yarp-devices-forcetorque` as per https://github.com/robotology/yarp-devices-forcetorque/issues/33


traversaro:
Thanks! In the past for other renames we never updated old release files, but indeed probably it make more send to update them.

traversaro:
Waiting for at least one job to be successful for merging.

PR #1198, opened 2022-06-29
Remove Unstable and LatestReleases jobs from releases/2022.05 CI

traversaro:
The only relevant job for this branch is `Default`, that tests the `2022.05.2.yaml` release.


traversaro:
Only `Default` job are correctly started, so we can directly merge, as nothing else changed.

PR #1199, opened 2022-06-29
Fix definition of GAZEBO_PLUGIN_PATH for Windows in setup scripts

traversaro:


Nicogene (commented):
So the plugins are loaded correctly both on linux and Windows if installed in `bin`?

traversaro:
> So the plugins are loaded correctly both on linux and Windows if installed in `bin`?

No, but Command Prompt and those PowerShell scripts are only used in Windows.

Nicogene (approved):
LGTM!

PR #1201, opened 2022-06-30
Do not show the ROBOTOLOGY_ENABLE_EVENT_DRIVEN option on Windows

traversaro:
The option is not supported on Windows as the repo https://github.com/robotology/event-driven is not supported on Windows, so do now expose the `ROBOTOLOGY_ENABLE_EVENT_DRIVEN` option on Windows to avoid that users enable it.


traversaro:
fyi @conconman @arrenglover

arrenglover:
thanks for letting me know. seems reasonable solution to me until we enable windows support.

PR #1202, opened 2022-06-30
Enable by default ROBOTOLOGY_USES_GAZEBO also on Windows

traversaro:
For a long time, `ROBOTOLOGY_USES_GAZEBO` was set to `OFF` on Windows as Gazebo was either not available or experimental. As Gazebo on Windows is now quite stable, leaving the option to OFF is creating more problems than it solves.


Nicogene (approved):
🎉

PR #1216, opened 2022-07-27
Switch to macos-11.0 and macos-latest in CI to handle macos-10.15 removal

traversaro:


traversaro:
I am afraid we will hit https://github.com/robotology/robotology-superbuild/issues/955, that we should have reported upstream back in time, instead of rushing now. -_-

traversaro:
Please ignore the Jammy failure @RiccardoGrieco @Nicogene, it is probably unrelated.

Nicogene (approved):

Issue #1187, opened 2022-06-27
Clarification on the usage of XDG_DATA_DIRS


xEnVrE:
Recently I have tried to setup a `robotology-superbuild` install on my system (Arch Linux with `5.18.7-arch1-1` kernel).

After sourcing the `build/install/share/robotology-superbuild/setup.sh` file, as per https://github.com/robotology/robotology-superbuild#configure-your-environment, I found out that my system could not run any UI dialog reliably anymore as many of the png files associated to icons could not be found. This was causing erratic crashes or strange behaviors.

After investigating a bit I found that the `setup.sh` file sets the `XDG_DATA_DIRS` as follows:

https://github.com/robotology/robotology-superbuild/blob/bab995bf57661af9f20d0907431910bac17096d1/cmake/template/setup.sh.in#L17

Unfortunately, in some systems, as the one I am using apparently, if the `XDG_DATA_DIRS` is not set or empty, a default value of `/usr/local/share/:/usr/share/` is assumed. However, when the `XDG_DATA_DIRS` is populated by the `setup.sh`, given that it is empty, the new value will just contain `<path_to_ROBOTOLOGY_SUPERBUILD_INSTALL_PREFIX>/share`, i.e. the reference to `/usr/local/share` and `/usr/share` is missing.

Of course, this seems to be very system-dependent and might be easily solved if the user exports `XDG_DATA_DIRS` correctly. However, on some systems this might create strange behaviors and it is difficult to track them down and associate them with the `robotology-superbuild` installation.

Wondering if you could suggest anything to mitigate the problem, e.g. adding even a simple disclaimer in the README.

Thanks

cc @traversaro

traversaro:
Thanks for reporting the issue so clearly. This is a indeed a pain point, in the `setup.sh` we should just append to `XDG_DATA_DIRS` if it is not defined, otherwise we should first define it as `/usr/local/share/:/usr/share/` and only later append, I will try to fix this issue.

For reference, this is the reason why this was added in the first place:
* https://github.com/robotology/robotology-superbuild/issues/622
* https://github.com/robotology/robotology-superbuild/pull/759

@HosameldinMohamed just a curiosity, did you ever experienced something like that?

xEnVrE:
Thanks for the explanation @traversaro

xEnVrE:
[...]

PR #1257, opened 2022-09-13
Fix Linux bug when XDG_DATA_DIRS is not defined

traversaro:
`XDG_DATA_DIRS` default value (at least on Linux system) when the variable is not set is `/usr/local/share/:/usr/share/`, see https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html . This means that the behaviour of the Linux application when `XDG_DATA_DIRS` is not set or when it is set to an empty string is not the same, as in the latter case the directories `/usr/local/share/:/usr/share/`, where most configuration files of system installed software are installed are ignored, resulting in software not working properly (see https://github.com/robotology/robotology-superbuild/issues/1187 for some examples).

On most system this issue was not appearing as some other system installed software was already properly defining `XDG_DATA_DIRS` with `/usr/local/share/:/usr/share/` in it, but on some system (such as barebones Docker images) that was not the case.

This PR fixes this problem by putting `/usr/local/share/:/usr/share/` in `XDG_DATA_DIRS` if `XDG_DATA_DIRS` was not defined, before appending any value to it. This logic is only applied on Linux and if conda is not used.

Fix https://github.com/robotology/robotology-superbuild/issues/1187 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
+### Fixed
+- Fixed problem that occured when robotology-superbuild's `setup.sh` was sourced on Linux system in which `XDG_DATA_DIRS` was not defined ().
+
 ## [2022.08.0] - 2022-09-02
 
 ### Changed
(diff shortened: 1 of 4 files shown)


traversaro:
@xEnVrE @HosameldinMohamed @RiccardoZuppetti This PR fixes the problem(s) discussed in https://github.com/robotology/robotology-superbuild/issues/1187 , feel free to approve if you like, thanks!

HosameldinMohamed (approved):
Cool!

Issue #1269, opened 2022-09-22
Windows/conda CI failure end of September 2022 


traversaro:
~~~
2022-09-22T04:48:55.8649732Z D:/a/robotology-superbuild/robotology-superbuild/build/install/include\BipedalLocomotion/TextLogging/Logger.h(12,10): fatal error C1083: Cannot open include file: 'spdlog/fmt/ostr.h': No such file or directory [D:\a\robotology-superbuild\robotology-superbuild\build\src\walking-controllers\src\RetargetingHelper\RetargetingHelper.vcxproj] [D:\a\robotology-superbuild\robotology-superbuild\build\walking-controllers.vcxproj]
~~~

traversaro:
I am unable to reproduce it locally, as the file is there. 

Last working job: https://github.com/robotology/robotology-superbuild/actions/runs/3097052794 .
First not-working job: https://github.com/robotology/robotology-superbuild/actions/runs/3104130137 .

traversaro:
Now the problem also happens in vcpkg builds. However, it does not appear on the `LatestReleases` job.

traversaro:
> I am unable to reproduce it locally, as the file is there.

I think that this was an error on my side: I tought the failure was in blf, but actually it was on walking-controllers. Probably I just tried to compile bipedal-locomotion-framework and that was working fine.

traversaro:
After understanding that the issue was actually in walking-controllers, the fix was trivial: https://github.com/robotology/walking-controllers/pull/125 .

traversaro:
The walking-controllers PR fixed the vcpkg, but now Windows/conda has another problem:
~~~
2022-09-26T08:52:36.8611720Z     Auto build dll exports
2022-09-26T08:52:37.0258116Z        Creating library D:/a/robotology-superbuild/robotology-superbuild/build/src/bipedal-locomotion-framework/lib/Release/BipedalLocomotionFrameworkPlanners.lib and object D:/a/robotology-superbuild/robotology-superbuild/build/src/bipedal-locomotion-framework/lib/Release/BipedalLocomotionFrameworkPlanners.exp
[...]

PR #1275, opened 2022-09-26
Only build qhull on Linux with apt dependencies

traversaro:
See https://github.com/robotology/robotology-superbuild/issues/1269#issuecomment-1257811559 . Basically, if we install PCL with a package manager, we now install only qhull, so we would have both qhull installed and the one compiled by the superbuild. On macOS|homebrew, Windows|vcpkg and conda there is no problem with the installed qhull, so we switch to just use the qhull installed by the package manager. 

To do that also on Ubuntu with apt dependencies, we would need to drop Ubuntu 20.04 , that I am afraid it will not happen anytime soon.

### DIFF
--- a/cmake/Buildbipedal-locomotion-framework.cmake
+++ b/cmake/Buildbipedal-locomotion-framework.cmake
@@ -24,16 +24,19 @@ set(bipedal-locomotion-framework_USES_CppAD OFF)
 
 if (ROBOTOLOGY_ENABLE_DYNAMICS_FULL_DEPS)
   find_or_build_package(manif QUIET)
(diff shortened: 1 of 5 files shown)


traversaro:
Hopefully this will finally fix https://github.com/robotology/robotology-superbuild/issues/1269 .

traversaro:
CI is finally fine.

traversaro:
As this is a fix to fix CI, I will merge, just mentioning people that may be interested in this a posteriori: @GiulioRomualdi @S-Dafarra @Nicogene @RiccardoGrieco .

PR #1280, opened 2022-10-01
Bump `icub-firmware-models@v1.26.0`

pattacini:
As per the title

### DIFF
--- a/releases/latest.releases.yaml
+++ b/releases/latest.releases.yaml
@@ -50,7 +50,7 @@ repositories:
   icub-models:
     type: git
     url: https://github.com/robotology/icub-models.git
-    version: v1.25.0
+    version: v1.26.0
   yarp-matlab-bindings:
     type: git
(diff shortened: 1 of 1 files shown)


traversaro (approved):

releases/latest.releases.yaml
```diff
@@ -50,7 +50,7 @@ repositories:
   icub-models:
     type: git
     url: https://github.com/robotology/icub-models.git
-    version: v1.25.0
```
traversaro, line 53: It seems that this is bumping `icub-models` instead of `icub-firmware-models`, as noted by the CI. @pattacini

pattacini: Ops! I'll fix it at once.
icub-firmware-models also was at v1.25.0 and this tricked me.

pattacini: Fixed in #1281.

PR #1281, opened 2022-10-01
Fix #1280 and bump `icub-firmware-models@v1.26.0`

pattacini:
This PR reverts the wrongly assigned version to `icub-models` (ref. #1280) and advances the version of `icub-firmware-models` (as it was originally intended).


traversaro (approved):
Ok for me, let's finish CI so we can check that everything works.

PR #1080, opened 2022-04-01
Create 2022.02.1 release

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1079 . Bump YCM to 0.13.1 to fix CMake 3.23 compatibility.

PR #1283, opened 2022-10-03
ICUB tag fixing deprecated OpenGL libraries

hypothe:
Set the default tag of ICUB repository to the 1.24.1, to avoid OpenGL errors raised at buildtime.

### DIFF
--- a/releases/2022.02.1.yaml
+++ b/releases/2022.02.1.yaml
@@ -34,7 +34,7 @@ repositories:
   ICUB:
     type: git
     url: https://github.com/robotology/icub-main.git
-    version: v1.24.0
+    version: v1.24.1
   ICUBcontrib:
     type: git
     url: https://github.com/robotology/icub-contrib-common.git


traversaro:
Hello @hypothe, thanks for the PR! Unfortunatly, existing releases are immutable. What we can done in this case is to do a new patch release and create 2022.02.2 (see https://github.com/robotology/robotology-superbuild/pull/1080). I guess you are from IIT, right? In this case we can quickly get aligned on IIT's Teams if it is ok for you, I pinged you there.

traversaro:
We aligned with @hypothe via teams. Basically what is happening is that the distro v2022.02.1 is not working correctly when it is finding in the system a YCM >= 0.14.0, as the problem that emerged due to https://github.com/robotology/ycm/pull/401 is still present in icub-main 1.24.0 . Bumping icub-main to v1.24.0 is solving the issue, so probably it make sense to release a small 2022.02.2 (@hypothe for your knoledge installing in the system YCM 0.13.1 [https://github.com/robotology/robotology-superbuild/blob/releases/2022.02/releases/2022.02.1.yaml#L29] would be an alternative option).

traversaro:
Superseded by https://github.com/robotology/robotology-superbuild/pull/1286 .

traversaro:
Superseded by https://github.com/robotology/robotology-superbuild/releases/tag/v2022.02.2 .

PR #1287, opened 2022-10-11
[CI]: Replaced `set-output`

pattacini:
`set-output` is now deprecated as per https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.


traversaro (approved):

traversaro:
Thanks!

PR #1288, opened 2022-10-12
Update master with info related to v2022.02.2 release

traversaro:


traversaro:
@Nicogene @pattacini if you want to re-generate the docs now the 2022.02.2 distro has been added.

pattacini:
We can wait for 2022.09.1, which will cover this case as well.

PR #1053, opened 2022-03-04
Fix icub-models python and conda support

traversaro:
There were a few bugs for which the Python helper library of `icub-models` was not compiled even if `ROBOTOLOGY_USES_PYTHON` was enabled.

PR #1331, opened 2023-01-25
Add ergocub-software to core profile

Nicogene:
This PR adds [`ergocub-software`](https://github.com/icub-tech-iit/ergocub-software) to the core profile.


cmake/ProjectsTagsUnstable.cmake
```diff
@@ -30,3 +30,4 @@ set_tag(icub-gazebo_TAG devel)
 set_tag(icub-gazebo-wholebody_TAG devel)
 set_tag(whole-body-controllers_TAG master)
 set_tag(gym-ignition_TAG v1.3.1)
+set_tag(ergocub-software_TAG master)
```
traversaro, line 33: If `master` is the default branch, we can avoid doing that for simplicity: 
```suggestion
```

traversaro:
Thanks, some additional comments for this PR: 
* Can you also update the docs in https://github.com/robotology/robotology-superbuild/blob/master/doc/environment-variables-configuration.md#generic-configuration ? 
* Can you also add `ergocub-software_activate.msh` and `ergocub-software_deactivate.msh` in `conda/multisheller` directory? You can take as an example the icub-models files that we removed in https://github.com/robotology/robotology-superbuild/pull/1053/files (for referece, those files were moved in https://github.com/conda-forge/icub-models-feedstock/tree/main/recipe

Not directly relaed to this PR: 
* in https://github.com/icub-tech-iit/ergocub-software#installation I think you need to also specify that `<install-prefix>/share` should be added to `GAZEBO_MODEL_PATH`, otherwise the meshes will not be found. I suspect you did not hit that as you always installed in the robotology-superbuild install prefix, that already defined that in the `setup.**` script
* Not sure you already looked into this, but perhaps we will need to also set `ROS_PACKAGE_PATH` and `AMENT_PREFIX_PATH` to use the ergoCub URDF model in rviz and rviz2?

traversaro (changes requested):
See comments.

traversaro:
For the record, I am a bit afraid of having so much values to append to env variables, as for example on some operating system the length of env variables is limited. Anyhow, this is definitely something that is not related to this PR.

Nicogene:
> Not sure you already looked into this, but perhaps we will need to also set ROS_PACKAGE_PATH and AMENT_PREFIX_PATH to use the ergoCub URDF model in rviz and rviz2?

I am not sure about this since I never tried to use `ergocub` with ros2, but since its configuration files are already `yarprobotinterface` compatible, should be easy to make it work with ros2 nws and check it.


I addressed to requested changes 👍🏻

conda/multisheller/ergocub-software_activate.msh
```diff
@@ -0,0 +1,27 @@
+if_(is_set("COMSPEC")).then_([
+	sys.list_append("YARP_DATA_DIRS", path.join(env("CONDA_PREFIX"), "Library\\share\\ergoCub")),
+	sys.list_append("ROS_PACKAGE_PATH", path.join(env("CONDA_PREFIX"), "Library\\share")),
+	sys.list_append("AMENT_PREFIX_PATH", path.join(env("CONDA_PREFIX"), "Library")),
```
traversaro, line 4: Just FYI, did does not do anything until you install the package.xml (for ROS1) or an appropriate file in ament's resource_index. Anyhow, they do not hurt for sure.

conda/multisheller/ergocub-software_deactivate.msh
```diff
@@ -0,0 +1,5 @@
+if_(is_set("COMSPEC")).then_([
+	sys.list_remove("YARP_DATA_DIRS", path.join(env("CONDA_PREFIX"), "Library\\share\\iCub"))
+]).else_([
+	sys.list_remove("YARP_DATA_DIRS", path.join(env("CONDA_PREFIX"), "share/iCub"))
+])
```
traversaro, line 5: This does not seem to match the variables that are modified in `ergocub-software_activate.msh` . The basic ideas of this files is that you modify the env variables to undo the changes you do in `ergocub-software_activate.msh`.

conda/multisheller/ergocub-software_activate.msh
```diff
@@ -0,0 +1,27 @@
...
+	sys.list_append("GAZEBO_MODEL_PATH", path.join(env("CONDA_PREFIX"), "Library\\share\\gazebo\\models")),
+	sys.list_append("GAZEBO_RESOURCE_PATH", path.join(env("CONDA_PREFIX"), "Library\\share\\gazebo\\worls"))
```
traversaro, line 6: I do not think you the ergocub-software package is installing anything in the `<install_prefix>/share/gazebo` folder, right? So probably we can just remove these lines?

conda/multisheller/ergocub-software_activate.msh
```diff
@@ -0,0 +1,27 @@
...
+]).else_([
+	sys.list_append("YARP_DATA_DIRS", path.join(env("CONDA_PREFIX"), "share/ergoCub")),
+	sys.list_append("ROS_PACKAGE_PATH", path.join(env("CONDA_PREFIX"), "share")),
+	sys.list_append("AMENT_PREFIX_PATH", env("CONDA_PREFIX")),
+	sys.list_append("GAZEBO_MODEL_PATH", path.join(env("CONDA_PREFIX"), "share/gazebo/models")),
+	sys.list_append("GAZEBO_RESOURCE_PATH", path.join(env("CONDA_PREFIX"), "share/gazebo/worls"))
```
traversaro, line 12: See https://github.com/robotology/robotology-superbuild/pull/1331/files#r1087670442 .

traversaro (changes requested):
See comments.

conda/multisheller/ergocub-software_deactivate.msh
Nicogene: You are right, I forgot to save the file and I committed the one I copied from icub-models, it should be fine now

conda/multisheller/ergocub-software_activate.msh
Nicogene: I removed them also in the `_deactivate` file

traversaro (approved):

traversaro (approved):

traversaro:
@Nicogene for the future please Squash&merge if there are a lot of fixup commits, thanks!

PR #1340, opened 2023-02-08
Enable the POS service by default when ICUB_HEAD in enabled

sgiraz:
**What's new:**

- This change enables the POS service by default when building the `robotology-superbuild` with ICUB_HEAD=ON. The `POS` service is mandatory to be able to read the absolute encoders mounted on `ergoCub's` hands. 

**Notes:**
- Other robots are not affected by this change


Nicogene (approved):

traversaro (approved):

traversaro:
Thanks @sgiraz !

Issue #1190, opened 2022-06-28
Update WSL2 documentation to cover WSLg


traversaro:
Reported by @FabioBergonti . Basically if a user is on Win11 and can use WSLg, we can skip setting the DISPLAY env variable and install a X server, see https://github.com/robotology/robotology-superbuild#wsl2 .

traversaro:
Ok. I probably forgot about this because I wanted to suggest IIT users to update to Windows 11 and WSLg before tackling this, but this proveed to be harder than I expected due to some internal policy. Anyhow, now WSLg is also available on Windows 10 (see https://devblogs.microsoft.com/commandline/the-windows-subsystem-for-linux-in-the-microsoft-store-is-now-generally-available-on-windows-10-and-11/), so probably we can just document WSLg workflow. 

@martinaxgloria just for reference, which Windows version did you got installed in your laptop? 

fyi @pattacini @davidelasagna

martinaxgloria:
Hi @traversaro, I had some problems with WSL on Windows 10, but I have updated to Windows 11 and now everything works correctly!

traversaro:
I double checked with @davidelasagna, and we (i.e. IIT) currently install on new laptops Windows 10 Version 22H2 (OS build 19045), while WSLg is available on Windows 10 since build 19044 (see https://learn.microsoft.com/en-us/windows/wsl/tutorials/gui-apps). So I think we can safely update the docs to just refer to WSLg.

PR #1343, opened 2023-02-14
Simplify WSL docs

traversaro:
In particular:
* Just document WSLg to run graphical applications as not it is available also on Windows 10
* Do not document anymore WSL1 as it is not really widely used, and it is not worth to document it (any user expert enough to run WSL1 will be able to configure its system appropriately)

Fix https://github.com/robotology/robotology-superbuild/issues/1190 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,24 +158,15 @@ If for any reason you do not want to use the provided `setup.sh` script and you 
 Please refer to [`doc/conda-forge.md`](doc/conda-forge.md) document for instruction on how to compile the superbuild from source using the conda-forge provided dependencies, in particular the [`Source Installation`](doc/conda-forge.md#source-installation) section.
 
 ## Windows Subsystem for Linux from source
+
 The [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl) (wsl)  lets developers run a GNU/Linux environment -- including most command-line tools, utilities, and applications -- directly on Windows, unmodified.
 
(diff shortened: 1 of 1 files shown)


traversaro:
For the sake of googlability, I will include here the information on how to use the robotology-superbuild with WSL1 :

### WSL1
With respect to WSL2, WSL1 uses the same IP address used by the Windows machine, so the amount of configuration and tweaks required are less.

#### Run graphical applications on WSL1
To run graphical applications on WSL, you need to install a X Server for Windows, that will be able to visualize the windows WSL-based applications, see https://www.howtogeek.com/261575/how-to-run-graphical-linux-desktop-applications-from-windows-10s-bash-shell/ for more info. For information of X Servers that can be installed on Windows, follow the docs in https://github.com/sirredbeard/Awesome-WSL#10-gui-apps .

#### Sanitize enviroment variables for WSL1
By default, the `PATH` enviroment variable in WSL will contain the path of the host Windows system, see https://github.com/microsoft/WSL/issues/1640 and https://github.com/microsoft/WSL/issues/1493. This can create problems,
as the CMake in WSL may find (incompatible) Windows CMake packages and try to use them, creating errors due to the compilation.
To avoid that, you can add the following line in the WSL `.bashrc` that filters all the Windows paths from the WSL's enviromental variables:
~~~
for var in $(env | awk {'FS="="} /\/mnt\//{print $1}'); do export ${var}=\"$(echo ${!var} | awk -v RS=: -v ORS=: '/\/mnt\// {next} {print $1}')\" ; done
~~~

PR #1364, opened 2023-03-16
Fix ergocub-software branch used in ProjectsTagsStable.cmake

traversaro:
@GiulioRomualdi noted that we did not updated to the ergocub-software 0.2.0 in the superbuild, even if it was release. By looking what is appened, it seems that the ergocub-software was pinned to v0.1.0 in `ProjectsTagsStable.cmake`: this prevents the update versions script to work correctly, as the update version scripts just checks the latest tag that occured in the stable branch.

The pin was added in https://github.com/robotology/robotology-superbuild/pull/1331, @Nicogene was that intentional or just a copy&paste error?


Nicogene:
> The pin was added in https://github.com/robotology/robotology-superbuild/pull/1331, @Nicogene was that intentional or just a copy&paste error?

It was a copy-paste error, my bad!

Nicogene (approved):

Issue #1378, opened 2023-04-05
Update casadi to 3.6.0 release


traversaro:
Currently we are relying on a casadi fork, in particular https://github.com/ami-iit/casadi/tree/support_osqp_0.6.2 (see https://github.com/robotology/robotology-superbuild/blob/v2023.02.1/cmake/Buildcasadi.cmake#L19). Let's see what we need to update to casadi 3.6.0 . @GiulioRomualdi @S-Dafarra

traversaro:
Local changes:
* ✅ Support for osqp 0.6 --> merged upstream in https://github.com/casadi/casadi/pull/2671
* ❌ Support for seeing casadi in pip list ---> the pr corresponding to https://github.com/ami-iit/casadi/pull/1, i.e. https://github.com/casadi/casadi/pull/2878 merged upstream in , but the PR corresponding to https://github.com/ami-iit/casadi/pull/2 , i.e. https://github.com/casadi/casadi/pull/2906, was never merged upsream

traversaro:
So we need to take the 3.6.0 release and add the patch https://github.com/casadi/casadi/pull/2906 on top of it. At that point, probably we can also add https://github.com/casadi/casadi/pull/2965 on top that it is useful to avoid qpoases-related incompatibility.

Issue #1388, opened 2023-04-16
CI Failure Mid April 2023 ROBOTOLOGY_ENABLE_DYNAMICS Stable and Unstable branches


traversaro:
The error is:

~~~
[...]

PR #1389, opened 2023-04-17
Buildcasadi: Enable CASADI_PYTHON_PIP_METADATA_INSTALL option

traversaro:
Fix comment reported in https://github.com/robotology/robotology-superbuild/issues/1378#issuecomment-1510960702 . In our fork of casadi 3.5.5 the option by default was ON (see https://github.com/ami-iit/casadi/pull/1/files#diff-75df003c8ca5bb396ddc7edb9b4b18685f77e287feaa44b6c9175d1a46364275R168), so we never did this at the superbuild level, however this is required now that we are using the upstream version after it was merged in https://github.com/casadi/casadi/pull/2878 .


diegoferigo (approved):
I didn't validate on my setup but the fix is pretty clear. Thanks @traversaro for addressing the issue this promptly.

traversaro:
Merging, `LatestReleases` pass fine, all other jobs fail due to https://github.com/robotology/robotology-superbuild/issues/1388, but we can merge.

Issue #916, opened 2021-11-06
Install robotology-superbuild natively on macOS arm64 (so-called "Apple Silicon")


traversaro:
In the future, the amount of members of the lab or external contributors that may be arm-based mac computers will increase. It will be ideal to get ready for that point, with instructions that explain how to install the robotology-superbuild on macOS arm64 natively. In the mean time, it probably could make sense to clearly document how to install the robotology-superbuild on [Rosetta2](https://support.apple.com/en-us/HT211861), but this is tracked in another issue: https://github.com/robotology/robotology-superbuild/issues/915 .

traversaro:
First of all, we should track which dependencies are available on macOS on arm64. Due to https://github.com/robotology/robotology-superbuild/issues/842, we are going to track this on conda, even if also Homebrew can install packages on macOS arm64 . We got the dependencies from https://github.com/robotology/robotology-superbuild/blob/8c81b4e310358db602c4fe834d19bf2aed8effa7/doc/conda-forge.md#create-an-environment-and-install-dependencies : 

[WIP] Open source dependencies (provided by conda-forge):
* ❌ ace 
* asio 
* boost 
* ✔️ eigen 
* gazebo 
* glew 
* glfw 
* gsl 
* ipopt 
* irrlicht 
* libjpeg-turbo 
* libmatio 
* libode 
* libxml2 
* nlohmann_json 
* opencv 
* pkg-config 
* portaudio 
* qt 
* sdl 
* sdl2 
* sqlite 
* tinyxml 
* spdlog

[WIP] Closed source dependencies:
* MATLAB/Simulink:

traversaro:
Graph of missing dependencies as of 2021/11/06 from https://conda-forge.org/status/#armosxaddition : 
![idyntree](https://user-images.githubusercontent.com/1857049/140608998-03cbdad7-8e6e-41ea-bf26-72d8a6e7b4e8.PNG)

traversaro:
> ❌ ace (https://github.com/conda-forge/ace-feedstock/issues/29)

As described in the related issue, getting ace to work on `osx-arm64` may not be trivial. As most of YARP (except for the `serial` device and the `shmem` device) works also without ace, we could just add a `ROBOTOLOGY_USES_ACE` to disable the use of ACE, that by default is `OFF` on Apple Silicon.

traversaro:
[...]

PR #1394, opened 2023-04-22
Fix installation docs skip-ace Mac

DanielePucci:
Because of https://github.com/robotology/robotology-superbuild/issues/916, one should use the option SKIP-ACE in the Yarp CMAKE. Also, one should not have ace in the mamba dependencies. This PR fixes the installation documentation in those respects


traversaro:
Thanks @DanielePucci, just a clarification: The modification you mention are necessary if you are using a macOS with recent **ARM**-based processor, while they are not necessary if you using macOS with an **Intel**-based processor.

DanielePucci:
Good point @traversaro, I tried to correct it with https://github.com/robotology/robotology-superbuild/commit/b0e5a8e13225e4be8ac23ce6a76837f93c297e36 and https://github.com/robotology/robotology-superbuild/commit/748005c8c354427d6ca384db424bd38936f2e2ab 

Take a look at them, in the case do not hesitate to integrate them

PR #1400, opened 2023-04-26
Add support for build binary installers for each release via conda constructor

traversaro:


traversaro:
This is stale, let's close unless there is more interest.

PR #1401, opened 2023-05-02
Clarification of Windows installation

S-Dafarra:
Before it was a bit too macOs centric and it was not clear what to do when using Windows.

This was a sort of regression of #1394 where it was not clear what to do when using Windows.


traversaro:
Thanks! cc @DanielePucci

PR #1403, opened 2023-05-09
Prevent robotologyGitStatus to close when run with double click

S-Dafarra:
On windows it gets pretty handy to doublle click on the script to run it. In this way, the window does not disappear.

Following https://askubuntu.com/questions/729904/how-to-check-if-script-was-executed-via-command-line-or-double-clicked I have used the value of ``SHLVL`` to detect if the script is launched from an existing terminal (and hence keeping the previous behavior) or from double click

I have tested it on Windows

### DIFF
--- a/scripts/robotologyGitStatus.sh
+++ b/scripts/robotologyGitStatus.sh
@@ -44,3 +44,8 @@ for subdir in ${subdirs}; do \
 done
 echo "--------------------------------------------"
 
+# Detect if running from terminal or from double-click. https://askubuntu.com/questions/729904/how-to-check-if-script-was-executed-via-command-line-or-double-clicked
+if (( SHLVL == 1 ))
+then 
+    read -p "Press enter to continue..."
+fi


traversaro (commented):
This result in the "Press enter to continue..." message appearing also in some non-double click case (for example, if you call the bash from a non-bash script, I think). In general I do not think it is a big problem, but for this particular script I am afraid this could happen in https://github.com/ami-iit/bipedal-locomotion-framework/blob/06927b8e243673f16e0dcb89938dc4ad9bf750b0/devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp#L1117 . If that blf code is not blocked by this modification, for me it is ok to merge.

traversaro:
@S-Dafarra @GiulioRomualdi is this something you want still in?

S-Dafarra:
> @S-Dafarra @GiulioRomualdi is this something you want still in?

To be honest, I did not manage to test it yet on Windows with the logger

traversaro:
@S-Dafarra I will close this to cleanup a bit the superbuild PRs, if it is something you want to work on, feel free to reopen.

PR #1404, opened 2023-05-10
Move documentation on installation of dependencies for additional CMake options in main readme (apt or conda-forge)

traversaro:
This puts a bit more complexity in the main README, but it improves the linearity of the documentation, as otherwise the user need to jump from one document to another to understand the commands needed to enable a given dependency.

This was changed after a feedback from @gabrielecaddeo


gabrielecaddeo:
I think it is far better in this way, thank you @traversaro !

PR #1407, opened 2023-05-12
Disable Homebrew from CI and do not support it anymore

traversaro:
As announced in https://github.com/robotology/robotology-superbuild/issues/842#issuecomment-1500326241, this PR removes Homebrew support and documentation.


traversaro:
It is time.

PR #1414, opened 2023-05-18
Add bayes-filters-lib if ROBOTOLOGY_ENABLE_DYNAMICS_FULL_DEPS is ON

### LINKED ISSUE #1410: Add bayes-filters-lib

GiulioRomualdi:
This closes #1410


traversaro:
@xEnVrE if you want/like to have a dedicated profile to keep bayes-filtes-lib, feel free to comment. In the meanwhile we can proceed with `ROBOTOLOGY_ENABLE_DYNAMICS_FULL_DEPS`.

traversaro:
I just realized that a step was missing in https://github.com/robotology/robotology-superbuild/pull/1415 , can you add the release in `releases/latest.releases.yaml` ?

cmake/Buildbayes-filters-lib.cmake
```diff
@@ -0,0 +1,15 @@
+# Copyright (C) Fondazione Istituto Italiano di Tecnologia
+# CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+
+include(YCMEPHelper)
+include(FindOrBuildPackage)
+
+ycm_ep_helper(bayes-filters-lib TYPE GIT
```
traversaro, line 7:
```suggestion
ycm_ep_helper(BayesFilters TYPE GIT
```

For historical reason (basically so that `find_or_build_package` works fine if we wanted to find the package outside the superbuild), we use CMake package names as identifiers for packages in the superbuild, so in this context we would need `BayesFilters` (see https://github.com/robotology/bayes-filters-lib/blob/0.10.0/CMakeLists.txt#L12). Note that we also need to change the file name and the calls to `find_or_build_package(bayes-filters-lib)` .

traversaro (commented):
Small comment. Furthermore, I do not know if you also want to change `    find_or_build_package(bipedal-locomotion-framework)` to add the dependency on `BayesFilters` and the option to enable it explicitly.

GiulioRomualdi:
> Small comment. Furthermore, I do not know if you also want to change ` find_or_build_package(bipedal-locomotion-framework)` to add the dependency on `BayesFilters` and the option to enable it explicitly.

It will be required after https://github.com/ami-iit/bipedal-locomotion-framework/pull/669 so that option does not exist for the time being. I think I can open a PR after https://github.com/ami-iit/bipedal-locomotion-framework/pull/669 gets merged

traversaro:
> @traversaro I think I applied the modification you suggested. Let me know if there is something else missing
> 
> > Small comment. Furthermore, I do not know if you also want to change ` find_or_build_package(bipedal-locomotion-framework)` to add the dependency on `BayesFilters` and the option to enable it explicitly.
> 
> It will be required after [ami-iit/bipedal-locomotion-framework#669](https://github.com/ami-iit/bipedal-locomotion-framework/pull/669) so that option does not exist for the time being. I think I can open a PR after [ami-iit/bipedal-locomotion-framework#669](https://github.com/ami-iit/bipedal-locomotion-framework/pull/669) gets merged

Ack, in any case even before the PR is merged, the option will be ignored, so we can already add it here, so we are ready once that PR is merged. Anyhow, it is not necessary to merge this PR.

traversaro:
Ubuntu 18.04 failures are unrelated, please ignore them, thanks!

Issue #1071, opened 2022-03-24
Enable h264 support on YARP in the superbuild?


traversaro:

traversaro:
xref: https://github.com/robotology/yarp/issues/2885

traversaro:
Fixed in https://github.com/robotology/robotology-superbuild/pull/1443 .

PR #1443, opened 2023-07-11
YARP: Enable compilation of bottle_compression_zlib, depthimage_compression_zlib and image_compression_ffmpeg portmonitors

traversaro:
This PR enable in YARP the CMake options to ensure that the following portmonitor are enabled:
* `bottle_compression_zlib`
* `depthimage_compression_zlib`
* `image_compression_ffmpeg`

These portmonitors are already used by the HSP team working on ergocub (fyi @andrearosasco) and can be useful for anyone sending big amount of data over a YARP network (@S-Dafarra @GiulioRomualdi @mebbaid @davidegorbani).

This PR also adds the required dependencis (ffmpeg and zlib) to the dependency of the superbuild. This was already discussed in https://github.com/robotology/robotology-superbuild/issues/1071, but we decided not to proceed as some devices in YARP (`ffmpeg_grabber` and `ffmpeg_writer`) were not compatible with ffmpeg 5. However, it turns out that `image_compression_ffmpeg` can be used and can be useful on its own, so I think it make sense to proceed with this change, even because it would permit some HSP/ergoCub demo to run with a stock superbuild.

The corresponding PR in the yarp-feedstock recipe is: https://github.com/conda-forge/yarp-feedstock/pull/31 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
       shell: bash -l {0}
       run: |
         # Dependencies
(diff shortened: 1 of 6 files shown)


traversaro:
CI is happy, we can address any comment in future PRs.

PR #1448, opened 2023-07-24
Add include fetchcontent in Fetchonnxruntimebinaries.cmake

GiulioRomualdi:
Without this, the compilation failed on ergocub-torso


traversaro (approved):

traversaro:
Thanks! Just a curiosity, which CMake version is installed in ergocub-torso ?

PR #1454, opened 2023-08-03
Update install-mambaforge.md

lornat75:
not sure it is correct, it solved the problem for me, please somebody review


traversaro:
Yes, I think it is correct!

Issue #1461, opened 2023-08-27
Regression in robotology iDynTree MATLAB binary package 


traversaro:
The CI for MATLAB binary package is now failing for MATLAB R2020b, R2021a, R2021b with error:
~~~
  /home/runner/work/_actions/matlab-actions/run-command/v1/dist/bin/glnxa64/run-matlab-command cd('/tmp/run_matlab_command-rjikJv');command_53d06710_a3a9_4cf4_a8d6_af6a6de004c6 
  
  	Sponsored License -- for use in support of a program or activity sponsored by MathWorks.
  	Not for government, commercial or other non-sponsored organizational use.
  
  {Error using iDynTree.Position (line 10)
  Invalid MEX-file '/home/runner/work/robotology-superbuild/robotology-superbuild/scripts/robotology-matlab/mex/iDynTreeMEX.mexa64':
  /usr/local/MATLAB/R2020b/bin/glnxa64/../../sys/os/glnxa64/libgfortran.so.5: version `GFORTRAN_10' not found (required by
  /home/runner/work/robotology-superbuild/robotology-superbuild/scripts/robotology-matlab/mex/../lib/././libspral.so)
  
  Error in command_53d06710_a3a9_4cf4_a8d6_af6a6de004c6 (line 3)
  pos = iDynTree.Position()
  } 
  exit status 1
~~~

The latest relase of MATLAB is instead working.

traversaro:
This is a "regression" due to https://github.com/conda-forge/ipopt-feedstock/pull/94, basically  the `libgfortran.so.5` shipped with MATLAB is too old for the gfortran used in conda-forge. While a possible solution is to rename `/usr/local/MATLAB/R2020b/bin/glnxa64/../../sys/os/glnxa64/libgfortran.so.5` to a different name to use the system's `libgfortran.so.5`, probably an easier option is just to drop support for older MATLAB releases.

traversaro:
Indeed, based on information from https://it.mathworks.com/support/requirements/previous-releases.html, since R2022a MATLAB supports gfortran 10, while previous releases supported gfortran 8.

PR #1462, opened 2023-08-27
Only test and document official compatibility of MATLAB binary packages with MATLAB >= R2022a

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1461 . In https://github.com/robotology/robotology-superbuild/pull/1462/commits/f8f31010f81f54d8fd14b5d28985d487c78e110e I also document this in the documentation.


traversaro:
No one complained, I guess >= R2022a compatibility is ok.

Issue #1472, opened 2023-09-10
Ubuntu Focal CI Failure Mid September 2023


traversaro:
Build are failing due to missing space:
~~~
2023-09-10T03:05:45.1578874Z       |                                                        ^~~~~
2023-09-10T03:05:45.1579514Z /__w/robotology-superbuild/robotology-superbuild/src/walking-controllers/src/WholeBodyControllers/src/QPInverseKinematics_osqp.cpp: At global scope:
2023-09-10T03:05:45.1580493Z /__w/robotology-superbuild/robotology-superbuild/src/walking-controllers/src/WholeBodyControllers/src/QPInverseKinematics_osqp.cpp:211:1: fatal error: error writing to /tmp/cclOSFV9.s: No space left on device
2023-09-10T03:05:45.1581001Z   211 | }
2023-09-10T03:05:45.1581182Z       | ^
2023-09-10T03:05:45.1581587Z compilation terminated.
2023-09-10T03:05:45.1582042Z [27/52] Building CXX object src/WholeBodyControllers/CMakeFiles/WholeBodyControllers.dir/src/QPInverseKinematics.cpp.o
2023-09-10T03:05:45.1582634Z FAILED: src/WholeBodyControllers/CMakeFiles/WholeBodyControllers.dir/src/QPInverseKinematics.cpp.o 
[...]

PR #1473, opened 2023-09-10
GitHub Actions: Free disk space to avoid out of disk space errors

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1472 by deleting unused directories (from https://github.com/apache/flink/blob/53702736382b80fa905494e69654061a84741e84/tools/azure-pipelines/free_disk_space.sh#L48C1-L53C40) in the host machine. Given that we had this problem in actions that are running in a contained, we had first to mount the folder that contained the files to delete in the container.


traversaro:
The fix worked fine.

Issue #1505, opened 2023-10-19
Windows CI Mid October 2023 - YARP compilation failure and CMake find wrong non-conda Python executable


traversaro:
See https://github.com/robotology/yarp/issues/3031 .

traversaro:
> See [robotology/yarp#3031](https://github.com/robotology/yarp/issues/3031) .

This is only one problem that emerged. A second problem that emerged is for some reason the wrong Python version is found by CMake, i.e. not the conda one, but one installed by the GitHub Action image:
~~~
2023-10-19T02:25:21.1169465Z -- Found Python3: C:/hostedtoolcache/windows/Python/3.12.0/x64/python3.exe (found version "3.12.0") found components: Interpreter 
2023-10-19T02:25:21.1616387Z Traceback (most recent call last):
2023-10-19T02:25:21.1617566Z   File "<string>", line 1, in <module>
2023-10-19T02:25:21.1618470Z ModuleNotFoundError: No module named 'distutils'
~~~

That eventually fails with a reason related to https://github.com/robotology/robotology-superbuild/issues/1238 and https://github.com/robotology/robotology-superbuild/issues/1371 . However, the problem is that the non-conda Python is found.

traversaro:
> See [robotology/yarp#3031](https://github.com/robotology/yarp/issues/3031) .

This was fixed, thanks @randaz81 ! Now th only remaining issue is the wrong python being found, I tried to tackle this in https://github.com/robotology/robotology-superbuild/pull/1506 but this introduces a macos regression.

PR #1506, opened 2023-10-22
Set default policies of CMake as the one of CMake 3.16

traversaro:
The main reason to do so is to ensure that https://cmake.org/cmake/help/latest/policy/CMP0094.html#policy:CMP0094 is set to NEW and fix the Python part of https://github.com/robotology/robotology-superbuild/issues/1505 .

If you are confused by the syntax, check `policy_max` argument in https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html .


traversaro:
For some reason, now macOS build fail. Interestingly, a similar PR in https://github.com/robotology/icub-models/pull/217 works fine also on macOS, and CMake finds the right Python.

traversaro:
Supeseded by https://github.com/robotology/robotology-superbuild/pull/1509 .

Issue #1371, opened 2023-03-17
Handling of the installation directory of some python subprojects fragile


traversaro:
The logic of installation of Python bindings for some subprojects is inherently fragile. The superbuild computes the installation location of python bindings in https://github.com/robotology/robotology-superbuild/blob/3d587be2b500b53c3a7c0fd0835a9886514d7f99/CMakeLists.txt#L39, via the code:
~~~py
from distutils import sysconfig; print(sysconfig.get_python_lib(1,0,prefix=''))
~~~

Then, for some subprojects we explicitly set that the python installation should be the one that we set (for example casadi https://github.com/robotology/robotology-superbuild/blob/3d587be2b500b53c3a7c0fd0835a9886514d7f99/cmake/Buildcasadi.cmake#L33), however for some others we just rely on the fact that the subprojets also compute their python installation directory via `from distutils import sysconfig; print(sysconfig.get_python_lib(1,0,prefix=''))`. However, this is fragile as the, for example, the directory returned by this code is different depending on whater we are running the code in the system python environment or in a virtual environment. We already had users realllllly confused by this.

 A better alternative is to use a strategy similar to the one that we used for casadi.

fyi @LoreMoretti

Issue #1238, opened 2022-08-24
Warning on Python 3.10/Ubuntu 22.04


traversaro:
~~~
-- Found Python3: /usr/bin/python3.10 (found version "3.10.4") found components: Interpreter
<string>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
<string>:1: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
<string>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
<string>:1: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
<string>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
<string>:1: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
~~~

traversaro:
Unfortunatly the `sysconfig` top module does not have the same functions of `distutils.sysconfig`, so the change is not as trivial as attempted in https://github.com/robotology/robotology-superbuild/pull/1245 . I tried a more complex version from https://github.com/krshrimali/pytorch/commit/361bce4824f37bee19ae31010b01b93a9bed8780, but unfortunatly it gives different results already in Ubuntu 22.04 :

~~~
traversaro@IITICUBLAP257:~$ python3 -c "from distutils import sysconfig; print(sysconfig.get_python_lib())"
<string>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
<string>:1: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
/usr/lib/python3/dist-packages
traversaro@IITICUBLAP257:~$ python3 -c "import os;import sysconfig;relative_site_packages = sysconfig.get_path('purelib').replace(sysconfig.get_path('data'), '').lstrip(os.path.sep);print(relative_site_packages)"
lib/python3.10/dist-packages
~~~

So, we need to think more about this. Probably the best thing to do is not to rely subprojects of the superbuild to use the same logic of the superbuild to detect the python installation directly, but we should specify via appopriate CMake options the Python install directory used by the superbuild.

traversaro:
[...]

PR #1508, opened 2023-10-22
Explicitly set Python install directory in subprojects 

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1371
Prerequisite to fix https://github.com/robotology/robotology-superbuild/issues/1238

This PR require the following PRs to be merged and the corresponding changes released:
* https://github.com/ami-iit/bipedal-locomotion-framework/pull/741
* https://github.com/robotology/icub-models/pull/218
* https://github.com/robotology/wearables/pull/197
* https://github.com/robotology/idyntree/pull/1124

So it may take some time to be ready.


traversaro:
> This PR require the following PRs to be merged and the corresponding changes released:

To be honest, it is not strictly required for the changes to be released: we can just merge this change, and if the changes are not present the behavior will fall back to the earlier behavior, that works fine for Python <= 3.11 .

Issue #1514, opened 2023-10-31
Generate conda packages on Windows failed


traversaro:
~~~
2023-10-31T09:57:43.3762929Z 
2023-10-31T09:57:43.3763253Z %SRC_DIR%>set "PIP_NO_BUILD_ISOLATION=False" 
2023-10-31T09:57:43.6316794Z CMake Warning (dev) at C:/Miniconda3/envs/test/conda-bld/icub-main_1698746024287/_build_env/Library/share/cmake-3.27/Modules/CMakeDependentOption.cmake:89 (message):
2023-10-31T09:57:43.6318522Z   Policy CMP0127 is not set: cmake_dependent_option() supports full Condition
2023-10-31T09:57:43.6319565Z   Syntax.  Run "cmake --help-policy CMP0127" for policy details.  Use the
2023-10-31T09:57:43.6320568Z   cmake_policy command to set the policy and suppress this warning.
2023-10-31T09:57:43.6321347Z Call Stack (most recent call first):
2023-10-31T09:57:43.6322577Z   C:/Miniconda3/envs/test/conda-bld/icub-main_1698746024287/_h_env/Library/share/yarp/cmake/YarpPlugin.cmake:346 (cmake_dependent_option)
2023-10-31T09:57:43.6324418Z   src/libraries/icubmod/imuFilter/CMakeLists.txt:5 (yarp_prepare_plugin)
2023-10-31T09:57:43.6325507Z This warning is for project developers.  Use -Wno-dev to suppress it.
2023-10-31T09:57:43.6326100Z 
2023-10-31T09:57:43.6326108Z 
2023-10-31T09:57:43.6326365Z %SRC_DIR%>set "PIP_NO_DEPENDENCIES=True" 
2023-10-31T09:57:43.6326757Z 
2023-10-31T09:57:43.6327004Z %SRC_DIR%>set "PIP_IGNORE_INSTALLED=True" 
2023-10-31T09:57:43.6327401Z 
2023-10-31T09:57:43.6327960Z %SRC_DIR%>set "PIP_CACHE_DIR=C:\Miniconda3\envs\test\conda-bld\icub-main_1698746024287\pip_cache" 
2023-10-31T09:57:43.6328761Z 
2023-10-31T09:57:43.6328939Z %SRC_DIR%>set "PIP_NO_INDEX=True" 
2023-10-31T09:57:43.6329320Z 
[...]

PR #1515, opened 2023-11-02
Also set Python_EXECUTABLE CMake option in generated conda recipes

traversaro:
To avoid CMake scripts of downstream projects calling `find_package(Python3)` that pick the wrong Python, in the generated conda recipes we explicitly set the `Python3_EXECUTABLE` option.

As some projects actually call `find_package(Python)`, we should also set the `Python_EXECUTABLE` option.

Fix https://github.com/robotology/robotology-superbuild/issues/1514 .


traversaro:
Manually run test job: https://github.com/robotology/robotology-superbuild/actions/runs/6730456393 .

traversaro:
> Manually run test job: https://github.com/robotology/robotology-superbuild/actions/runs/6730456393 .

That works, `robotology-distro

PR #1535, opened 2023-11-20
Add R2023b to matlab-one-line-install-test CI job

traversaro:


traversaro:
I could not replicate the problem reported by @DavideBarbone . Tomorrow we can debug together.

traversaro:
It turns out that the problem was https://github.com/robotology/idyntree/issues/985, that affect all users of MATLAB that install all the toolbox, so it is independent from the MATLAB version.

Issue #1541, opened 2023-11-27
conda macos CI failure end of November 2023


traversaro:
~~~
2023-11-27T02:18:07.2233210Z }
2023-11-27T02:18:07.2279670Z loading initial cache file /Users/runner/work/robotology-superbuild/robotology-superbuild/.ci/initial-cache.gh.cmake
2023-11-27T02:18:08.0119710Z -- The C compiler identification is Clang 15.0.7
2023-11-27T02:18:09.2030750Z -- Detecting C compiler ABI info
2023-11-27T02:18:09.6929350Z -- Detecting C compiler ABI info - done
2023-11-27T02:18:09.7024160Z -- Check for working C compiler: /Users/runner/miniconda3/envs/test/bin/x86_64-apple-darwin13.4.0-clang - skipped
2023-11-27T02:18:09.7028290Z -- Detecting C compile features
2023-11-27T02:18:09.7036350Z -- Detecting C compile features - done
2023-11-27T02:18:09.9388660Z -- The CXX compiler identification is Clang 15.0.7
2023-11-27T02:18:10.0545950Z -- Detecting CXX compiler ABI info
2023-11-27T02:18:10.5563550Z -- Detecting CXX compiler ABI info - done
2023-11-27T02:18:10.5662290Z -- Check for working CXX compiler: /Users/runner/miniconda3/envs/test/bin/x86_64-apple-darwin13.4.0-clang++ - skipped
2023-11-27T02:18:10.5665760Z -- Detecting CXX compile features
2023-11-27T02:18:10.5674900Z -- Detecting CXX compile features - done
2023-11-27T02:18:13.6451380Z CMake Error at /Users/runner/miniconda3/envs/test/share/cmake-3.27/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
2023-11-27T02:18:13.6558070Z   Could NOT find Python3 (missing: Python3_LIBRARIES Development
2023-11-27T02:18:13.6559910Z -- Configuring incomplete, errors occurred!
2023-11-27T02:18:13.6793910Z   Development.Embed) (found version "3.12.0")
2023-11-27T02:18:13.6798830Z Call Stack (most recent call first):
2023-11-27T02:18:13.6804470Z   /Users/runner/miniconda3/envs/test/share/cmake-3.27/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
[...]

Issue #1510, opened 2023-10-29
Understand why setting CMP0094 to NEW results in a failure on APPLE


traversaro:
With policy_max set to 3.16, the macos config fails with:
~~~
loading initial cache file /Users/runner/work/robotology-superbuild/robotology-superbuild/.ci/initial-cache.gh.cmake
-- The C compiler identification is Clang 15.0.7
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Users/runner/miniconda3/envs/test/bin/x86_64-apple-darwin13.4.0-clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- The CXX compiler identification is Clang 15.0.7
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Users/runner/miniconda3/envs/test/bin/x86_64-apple-darwin13.4.0-clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /Users/runner/miniconda3/envs/test/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Python3 (missing: Python3_LIBRARIES Development
  Development.Embed) (found version "3.11.6")
Call Stack (most recent call first):
  /Users/runner/miniconda3/envs/test/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  /Users/runner/miniconda3/envs/test/share/cmake-3.26/Modules/FindPython/Support.cmake:3766 (find_package_handle_standard_args)
  /Users/runner/miniconda3/envs/test/share/cmake-3.26/Modules/FindPython3.cmake:551 (include)
  CMakeLists.txt:36 (find_package)


-- Configuring incomplete, errors occurred!
~~~

traversaro:
See also https://github.com/robotology/robotology-superbuild/actions/runs/6684105486/job/18161018643?pr=1509 : 
~~~
-- Check for working CXX compiler: /Users/runner/miniconda3/envs/test/bin/x86_64-apple-darwin13.4.0-clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /Users/runner/miniconda3/envs/test/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
[...]

Issue #1536, opened 2023-11-24
Unstable CI failure End of November 2023


traversaro:
~~~
2023-11-24T02:30:36.0548549Z [92/450] Building CXX object src/libraries/icubmod/bcbBattery/CMakeFiles/bcbBattery.dir/yarp_plugin_icubmod_bcbBattery.cpp.o
2023-11-24T02:30:36.0550512Z FAILED: src/libraries/icubmod/bcbBattery/CMakeFiles/bcbBattery.dir/yarp_plugin_icubmod_bcbBattery.cpp.o 
2023-11-24T02:30:36.0557289Z /usr/bin/c++  -DbcbBattery_EXPORTS -I/__w/robotology-superbuild/robotology-superbuild/src/ICUB/src/libraries/icubmod/bcbBattery -isystem /__w/robotology-superbuild/robotology-superbuild/build/install/include -g -fPIC   -std=c++17 -MD -MT src/libraries/icubmod/bcbBattery/CMakeFiles/bcbBattery.dir/yarp_plugin_icubmod_bcbBattery.cpp.o -MF src/libraries/icubmod/bcbBattery/CMakeFiles/bcbBattery.dir/yarp_plugin_icubmod_bcbBattery.cpp.o.d -o src/libraries/icubmod/bcbBattery/CMakeFiles/bcbBattery.dir/yarp_plugin_icubmod_bcbBattery.cpp.o -c src/libraries/icubmod/bcbBattery/yarp_plugin_icubmod_bcbBattery.cpp
2023-11-24T02:30:36.0563284Z In file included from src/libraries/icubmod/bcbBattery/yarp_plugin_icubmod_bcbBattery.cpp:10:
2023-11-24T02:30:36.0566241Z /__w/robotology-superbuild/robotology-superbuild/src/ICUB/src/libraries/icubmod/bcbBattery/bcbBattery.h:48:77: error: 'BATTERY_OK_STANBY' is not a member of 'yarp::dev::IBattery::Battery_status'
2023-11-24T02:30:36.0568640Z    48 |     IBattery::Battery_status     battery_status = IBattery::Battery_status::BATTERY_OK_STANBY;
2023-11-24T02:30:36.0569761Z       |                                                                             ^~~~~~~~~~~~~~~~~
[...]

PR #1546, opened 2023-11-27
CMake: Specify enabled languages in project instead of calling enable_language

traversaro:
I have no idea why (probably some hard to detect CMake bug) but in some cases:

~~~cmake
project(robotology-superbuild NONE)
enable_language(C)
enable_language(CXX)
~~~

make the first call to `find_package(Python3 COMPONENTS Interpreter Development REQUIRED)` fail, while:

~~~cmake
project(robotology-superbuild LANGUAGES C CXX)
~~~

works fine (beside being a more usual form).

Fix https://github.com/robotology/robotology-superbuild/issues/1541 .
Fix https://github.com/robotology/robotology-superbuild/issues/1510 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,21 +3,15 @@
 # CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
 
 cmake_minimum_required(VERSION 3.12)
-project(robotology-superbuild NONE)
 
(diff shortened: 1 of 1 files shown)


traversaro:
Other failures are due to https://github.com/robotology/robotology-superbuild/issues/1536, merging.

PR #1552, opened 2023-11-28
Set policy_max in cmake_minimum_required to 3.27

traversaro:
This should ensure all the policy are set to their `NEW` value.


traversaro:
This does not fix any failure, but it is anyhow beneficial, let's merge.

PR #1553, opened 2023-11-28
Debug macOs failure

traversaro:
Just debugging https://github.com/robotology/robotology-superbuild/issues/1541 .


traversaro:
I am unable to reproduce the failure with action-tmate .

PR #1539, opened 2023-11-24
ergocub-software: compile ergoCubEmotions by default if OpenCV >= 4.5.2

Nicogene:
It fixes #1513

PR #1558, opened 2023-12-15
Add depedency of ergocub-software on YARP

traversaro:
Required since https://github.com/icub-tech-iit/ergocub-software/pull/196 .


Nicogene:
There was this PR
- https://github.com/robotology/robotology-superbuild/pull/1539 
But I forgot to revive it after we made the change in `ergocub-software`

I will rebase it and I think we can merge it

PR #1562, opened 2023-12-28
Update build-docker-images.yml

pattacini:
Fix the problem due to a breaking change introduced with `tibdex/github-app-token@v2`.

See updated [documentation](https://github.com/tibdex/github-app-token/blob/main/action.yml).


Nicogene (approved):

traversaro:
Thanks!

PR #1598, opened 2024-02-13
Use GLFW3 all caps to avoid issues on Windows

### LINKED ISSUE #1597: IDYNTREE_USES_IRRLICHT is not properly set in Windows even if irrlicht and glfw3 are present in the system
During a recent demo we noticed that when recompiling the ``superbuild`` on a Windows machine, the ``IDYNTREE_USES_IRRLICHT`` was getting disabled even if ``irrlicht`` and ``glfw3`` are properly found in the system.

I tried to replicate it in my system and I noticed the same.

A perfect occasion to use the CMake debugging functionality of Visual Studio.

I blocked the execution at this line https://github.com/robotology/robotology-superbuild/blob/93504fbd876dbbab3ec7d2a7417942679a9827d7/cmake/BuildiDynTree.cmake#L29

Here I noted that the ``GLFW3_FOUND`` variable is set to true

image

On the other nad, inside the ``ycm_ep_helper`` function, in the arguments list, I see that ``-DIDYNTREE_USES_IRRLICHT`` is empty, i.e.
```
-DIDYNTREE_USES_IRRLICHT:BOOL=
```
image
while for example `-DIDYNTREE_USES_PYTHON:BOOL=True`

I believe that the problem is the case. In Windows, the variable is all capital, i.e. ``GLFW3_FOUND``, while we check the content of ``glfw3_FOUND``.

I think that the problem can be fixed by using 
```
find_package(GLFW3 QUIET)
```
instead

S-Dafarra:
Fixes: https://github.com/robotology/robotology-superbuild/issues/1597

### DIFF
--- a/cmake/BuildiDynTree.cmake
+++ b/cmake/BuildiDynTree.cmake
@@ -24,7 +24,7 @@ endif()
 # remove once irrlicht+sdl is available on vcpkg,
 # and https://github.com/robotology/robotology-superbuild-dependencies-vcpkg
 # is modified to includ it
-find_package(glfw3 QUIET)
+find_package(GLFW3 QUIET)
 
 ycm_ep_helper(iDynTree TYPE GIT
               STYLE GITHUB
@@ -34,7 +34,7 @@ ycm_ep_helper(iDynTree TYPE GIT
               FOLDER src
               CMAKE_ARGS -DIDYNTREE_USES_IPOPT:BOOL=ON
                          -DIDYNTREE_USES_OSQPEIGEN:BOOL=ON
-                         -DIDYNTREE_USES_IRRLICHT:BOOL=${glfw3_FOUND}
+                         -DIDYNTREE_USES_IRRLICHT:BOOL=${GLFW3_FOUND}
                          -DIDYNTREE_USES_ASSIMP:BOOL=ON
                          -DCMAKE_DISABLE_FIND_PACKAGE_YARP:BOOL=ON
                          -DIDYNTREE_USES_YARP:BOOL=OFF
--- a/cmake/BuildiDynTree.cmake
+++ b/cmake/BuildiDynTree.cmake
@@ -24,7 +24,15 @@ endif()
 # remove once irrlicht+sdl is available on vcpkg,
 # and https://github.com/robotology/robotology-superbuild-dependencies-vcpkg
 # is modified to includ it
-find_package(GLFW3 QUIET)
+find_package(glfw3 QUIET)
+
+# See https://github.com/robotology/robotology-superbuild/pull/1598#issuecomment-1942172913
+if(glfw3_FOUND OR GLFW3_FOUND)
+  set(IDYNTREE_USES_IRRLICHT ON)
+else()
+  set(IDYNTREE_USES_IRRLICHT OFF)
+endif()
+
 
 ycm_ep_helper(iDynTree TYPE GIT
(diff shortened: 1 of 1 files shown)


traversaro:
Thanks! 

A bit of background, `FindGLFW3` needs to be capital here as there is https://github.com/robotology/ycm/blob/master/find-modules/FindGLFW3.cmake in YCM. My error was to follow official glfw3 docs (https://www.glfw.org/docs/3.3/build_guide.html#build_link_cmake_package), that instead refer to `glfw3`, as glfw3 itself ships a `glfw3Config.cmake` (see for example https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Fglfw-3.3.9-hcfcfb64_0.conda). To avoid further confusion like this, I would like to deprecate the `FindGLFW3` in YCM, and instead use just the upstream glew3Config.cmake file. However, this would not work if we did `find_package(GLFW3)` here, as `glew3Config.cmake` will only be found by `find_package(glfw3)` calls, not `find_package(GLFW3)` calls (this is due to the fact that the file is named `glfw3Config.cmake`, if it was named `glfw3-config.cmake` any call to `find_package(glfw3)` , `find_package(GLFW3)` or `find_package(GlFw3)`  would always find `glfw3-config.cmake`, see https://cmake.org/cmake/help/latest/command/find_package.html#search-modes).

So, to be future friendly, I would stick to `find_package(glfw3)`, and just enable irrlicht if either `glfw3_FOUND` or `GLFW3_FOUND` are true, what do you think?

cmake/BuildiDynTree.cmake
```diff
@@ -24,7 +24,7 @@ endif()
 # remove once irrlicht+sdl is available on vcpkg,
 # and https://github.com/robotology/robotology-superbuild-dependencies-vcpkg
 # is modified to includ it
-find_package(glfw3 QUIET)
+find_package(GLFW3 QUIET)
```
traversaro, line 27:
```suggestion
find_package(glfw3 QUIET)
```

See PR discussion.

cmake/BuildiDynTree.cmake
```diff
@@ -34,7 +34,7 @@ ycm_ep_helper(iDynTree TYPE GIT
               FOLDER src
               CMAKE_ARGS -DIDYNTREE_USES_IPOPT:BOOL=ON
                          -DIDYNTREE_USES_OSQPEIGEN:BOOL=ON
-                         -DIDYNTREE_USES_IRRLICHT:BOOL=${glfw3_FOUND}
+                         -DIDYNTREE_USES_IRRLICHT:BOOL=${GLFW3_FOUND}
```
traversaro, line 37:
```suggestion
                         -DIDYNTREE_USES_IRRLICHT:BOOL=${IDYNTREE_USES_IRRLICHT}
```

traversaro:
By the way, interestingly we already worked on something related to this:
* https://github.com/robotology/ycm/pull/135
* https://github.com/robotology/ycm/issues/133

S-Dafarra:
Yeah it rang a bell also to me: https://github.com/ami-iit/yarp-device-openxrheadset/commit/a1eeaed7e6ebb9b28076f8839535be4e8c660b9d

Your solution sounds good!

traversaro:
Superseded by https://github.com/robotology/robotology-superbuild/pull/1599 .

PR #1599, opened 2024-02-13
Do not use find_package(glfw3) to detect if irrlicht+sdl is available

traversaro:
Solution for https://github.com/robotology/robotology-superbuild/issues/1597, alternative to https://github.com/robotology/robotology-superbuild/pull/1598 . The main problem of  https://github.com/robotology/robotology-superbuild/pull/1598  is that it would raise a deprecation warning after https://github.com/robotology/ycm/pull/441 is merged. Instead, we can already avoid to use the `FindGLFW3` script by passing `NO_MODULE` to `find_package(glfw3)`.

### DIFF
--- a/cmake/BuildiDynTree.cmake
+++ b/cmake/BuildiDynTree.cmake
@@ -23,8 +23,13 @@ endif()
 # Hack for disabling IDYNTREE_USES_IRRLICHT on vcpkg,
 # remove once irrlicht+sdl is available on vcpkg,
 # and https://github.com/robotology/robotology-superbuild-dependencies-vcpkg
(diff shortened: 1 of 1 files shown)


S-Dafarra (approved):

traversaro:
Cool, somehow `find_package(GLFW3)` failed on vcpkg, but `find_package(glfw3 NO_MODULE)` works fine and find `glfw3`, meaning that it is not a good way to detect if use or not irrlicht. Probably for the sake of the workaround, we can just enable irrlicht if conda is used.

S-Dafarra (approved):

Issue #1596, opened 2024-02-13
Propagate CMAKE_EXPORT_COMPILE_COMMANDS option?


traversaro:
Perhaps it could make sense that if `CMAKE_EXPORT_COMPILE_COMMANDS` is set to `ON` in the robotology-superbuild, we also set it to on in all the subprojects? 

@LoreMoretti

traversaro:
I saw that in https://cmake.org/cmake/help/v3.27/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html that:
> This is initialized by the [CMAKE_EXPORT_COMPILE_COMMANDS](https://cmake.org/cmake/help/v3.27/envvar/CMAKE_EXPORT_COMPILE_COMMANDS.html#envvar:CMAKE_EXPORT_COMPILE_COMMANDS) environment variable, and initializes the [EXPORT_COMPILE_COMMANDS](https://cmake.org/cmake/help/v3.27/prop_tgt/EXPORT_COMPILE_COMMANDS.html#prop_tgt:EXPORT_COMPILE_COMMANDS) target property for all targets.

Can't we just tell people to set `export CMAKE_EXPORT_COMPILE_COMMANDS=1` in their `.bashrc` or in their conda environment? I know that we told people not to put stuff in their .bashrc, but this seems quite a generic think that it could make sense to have in their .bashrc. In this way, we do not need to modify our code.

traversaro:
A second point is how to ensure that the multiple `compile_commands.json` of the subproject of the superbuild can be found by VSCode. I guess colcon and catkin workspaces have exactly the same problems, and indeed I found some related code for catkin: https://gist.github.com/AlexisTM/a6bb5ee488611e15114aa190b6f79404 . Apparently colcon does this automatically, see https://colcon.readthedocs.io/en/released/user/how-to.html#cmake-packages-generating-compile-commands-json . The related code is in https://github.com/colcon/colcon-cmake/blob/1509c77ecb417dd0b47e4d54b73833fa5e4c1baa/colcon_cmake/event_handler/compile_commands.py#L87-L104 .

traversaro:
I renamed the issue to track the actual problem we want to face.

traversaro:
Other related issues in cmake tools for vs:
* https://github.com/microsoft/vscode-cmake-tools/issues/915
* https://github.com/microsoft/vscode-cmake-tools/pull/2029
* https://github.com/microsoft/vscode-cmake-tools/issues/1120

[...]

PR #1600, opened 2024-02-14
Add vscode settings

giotherobot:
In this PR I add the VSCode `settings.json` and the CMake options needed to enable the intellisense integration for the superbuild.

See https://github.com/robotology/robotology-superbuild/issues/1596 for the discussion.

### DIFF
--- a/cmake/RobotologySuperbuildOptions.cmake
+++ b/cmake/RobotologySuperbuildOptions.cmake
@@ -3,10 +3,8 @@
 
 set(YCM_USE_CMAKE_PROPOSED TRUE CACHE BOOL "Use files including unmerged cmake patches")
 
-  set(CMAKE_EXPORT_COMPILE_COMMANDS "ON"
-    CACHE BOOL "Enable/Disable output of compile commands during generation."
-    )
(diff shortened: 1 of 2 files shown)


cmake/RobotologySuperbuildOptions.cmake
```diff
@@ -3,6 +3,8 @@
 
 set(YCM_USE_CMAKE_PROPOSED TRUE CACHE BOOL "Use files including unmerged cmake patches")
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
traversaro, line 6: Let's define the variable as an option, otherwise users can't override this setting if they like (I used the same style used in https://github.com/Kitware/CMake/blob/d6f32880c6b36e614a887b641be8103560c1e9e8/Modules/CMakeGenericSystem.cmake#L75C1-L78C50):

```suggestion
  set(CMAKE_EXPORT_COMPILE_COMMANDS "ON"
    CACHE BOOL "Enable/Disable output of compile commands during generation."
    )
  mark_as_advanced(CMAKE_EXPORT_COMPILE_COMMANDS)
```

traversaro (approved):

Issue #1602, opened 2024-02-22
Install both robotology-superbuild from source and ros noetic in conda result in dependency error (February 2024)


traversaro:
We experienced this today with @marcoforleo @Gianlucamilani @dariosortino .

~~~
mamba create -n testros -c conda-forge -c robostack-staging ace asio assimp boost eigen freetype gazebo glew glfw glm "graphviz==8.*" gsl ipopt irrlicht libjpeg-turbo libmatio libode libxml2 nlohmann_json pcl opencv portaudio qt-main sdl sdl2 sqlite tinyxml tinyxml2 spdlog lua soxr qhull cmake compilers make ninja pkg-config tomlplusplus libzlib ffmpeg onnxruntime-cpp ros-noetic-desktop
~~~

reports this error:
~~~
traversaro@IITBMP014LW012:~$ mamba create -n testros -c conda-forge -c robostack-staging ace asio assimp boost eigen freetype gazebo glew glfw glm "graphviz==8.*" gsl ipopt irrlicht libjpeg-turbo libmatio libode libxml2 nlohmann_json pcl opencv portaudio qt-main sdl sdl2 sqlite tinyxml tinyxml2 spdlog lua soxr qhull cmake compilers make ninja pkg-config tomlplusplus libzlib ffmpeg onnxruntime-cpp ros-noetic-desktop

Looking for: ['ace', 'asio', 'assimp', 'boost', 'eigen', 'freetype', 'gazebo', 'glew', 'glfw', 'glm', 'graphviz=8', 'gsl', 'ipopt', 'irrlicht', 'libjpeg-turbo', 'libmatio', 'libode', 'libxml2', 'nlohmann_json', 'pcl', 'opencv', 'portaudio', 'qt-main', 'sdl', 'sdl2', 'sqlite', 'tinyxml', 'tinyxml2', 'spdlog', 'lua', 'soxr', 'qhull', 'cmake', 'compilers', 'make', 'ninja', 'pkg-config', 'tomlplusplus', 'libzlib', 'ffmpeg', 'onnxruntime-cpp', 'ros-noetic-desktop']

robostack-staging/linux-64                                    No change
robostack-staging/noarch                                      No change
conda-forge/noarch                                  13.6MB @   5.7MB/s  2.6s
conda-forge/linux-64                                32.6MB @   8.2MB/s  4.3s
Could not solve for environment specs
The following packages are incompatible
├─ gazebo is installable with the potential options
[...]

PR #1538, opened 2023-11-24
Pin graphviz to 8

traversaro:
Workaround for https://github.com/robotology/robotology-superbuild/issues/1537 .

Issue #1537, opened 2023-11-24
conda-forge CI failure End of November 2023 due to graphviz update


traversaro:
Failure on Windows, macOS, Linux:

~~~
2023-11-24T02:16:57.2275069Z [17/1232] Building CXX object extern/qgv/CMakeFiles/YARP_priv_qgvcore.dir/qgv/QGVCore/private/QGVCore.cpp.o
2023-11-24T02:16:57.2276505Z FAILED: extern/qgv/CMakeFiles/YARP_priv_qgvcore.dir/qgv/QGVCore/private/QGVCore.cpp.o 
[...]

PR #1603, opened 2024-02-22
Remove pin of graphviz to 8 and disable yarpviz compilation on conda until YARP 3.10.0 is released

traversaro:
The pin was added in https://github.com/robotology/robotology-superbuild/pull/1538 as a workaround for https://github.com/robotology/robotology-superbuild/issues/1537, as the proper fixes have been released, we should be able to just drop the pin and use graphviz 9, to avoid incompatibility with libraries that actually use graphviz 9.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
       shell: bash -l {0}
       run: |
         # Dependencies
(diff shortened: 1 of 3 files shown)


traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1602 .

traversaro:
> the proper fixes have been released

This is not true. https://github.com/robotology/yarp/pull/3056 was merged, but was never released, and `yarp-3.9` branch and yarp 3.9.0 tag still do not compile with graphviz 9. As graphviz is just used for yarpviz compilation, and as I think that ROS compatibility (see https://github.com/robotology/robotology-superbuild/issues/1602) is more important, I think we can change the workaround to actually disable yarpviz on conda, and re-enable it once YARP 3.10 is released.

traversaro:
Indeed, the build still fails, let's disable yarpviz compilation on conda instead.

traversaro:
> let's disable yarpviz compilation on conda instead.

Done in https://github.com/robotology/robotology-superbuild/pull/1603/commits/03f7933f104fba82e53a4af5cc6d28feeb34e9dd .

PR #1620, opened 2024-03-22
Update install-mambaforge.md

xela-95:
I updated the sentence explaining how to run the `conda init` and `mamba init` commands, for increased clarity. By reading this quickly it can be erroneously intended to put that commands directly in the `.bashrc` instead of executing them on the shell (these commands in turn will update the `.bashrc`).


traversaro:
Thanks! This hopefully should reduce the confusions such as the one experienced by @LoreMoretti and @Gio-DS, thanks! We can think of also removing altogether the `.bashrc` reference, but that hopefully should be more clear.

PR #1621, opened 2024-03-22
Fix settings.json cmake merged commands location

giotherobot:
I was having problems with the mergedCompileCommands setting: the expression was not evaluated correctly and so the file was generated in the wrong folder.

This should correct the behaviour.


traversaro:
fyi @LoreMoretti @xela-95 , if there are more problems feel free to report.

PR #1624, opened 2024-03-26
Add missing dependencies for robot-log-visualizer

giotherobot:
We were missing some dependencies needed for the robot-log-visualizer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,7 +177,7 @@ jobs:
 
     # Skip on Apple Silicon https://github.com/robotology/robotology-superbuild/issues/1625
     - name: Check python metadata of installed python packages are coherent [Conda]
-      if: !contains(matrix.os, ' macos-14')
+      if: matrix.os != 'macos-14'
       shell: bash -l {0}
       run: |
         source ./b/install/share/robotology-superbuild/setup.sh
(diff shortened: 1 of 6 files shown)


traversaro:
Thanks! For reference `qtpy` is required by `pyqtconsole`, and `pyqtwebengine` is instead a proper robot-log-visualizer dependency.

traversaro:
The `pip check` that I added reported another problem:

~~~
DEPRECATION: yarp 3.9.0+9-20240316.1+gitf0e6750c4 has a non-standard version number. pip 24.1 will enforce this behaviour change. A possible replacement is to upgrade to a newer version of yarp or contact the author to suggest that they release a version with a conforming version number. Discussion can be found at https://github.com/pypa/pip/issues/12063
meshcat 0.3.2 requires pyngrok, which is not installed.
Error: Process completed with exit code 1.
~~~

pyngrok is not available in apt (see https://repology.org/project/python:pyngrok/versions), so probably we need to add it here. Not sure if it make sense to block the PR on that, but it is relatively easy, we can do that.

traversaro:
Cool, now only Apple Silicon builds fail with:
~~~
 robot-log-visualizer 0.8.1.dev2 requires pyqt5, which is not installed.
robot-log-visualizer 0.8.1.dev2 requires pyqtwebengine, which is not installed.
~~~

not worth enough to block further the PR, let's open a new issue: https://github.com/robotology/robotology-superbuild/issues/1625 .

traversaro (approved):

PR #1547, opened 2023-11-27
Add yarp-ros repo with Unstable branches and compile again against YARP master in Unstable branches

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1536 .

Changes necessary to support modern YARP:
* Drop support for Debian 10 Buster and 11 Bullseye (and hence of pc104-based iCubs, see https://icub-tech-iit.github.io/documentation/icub_operating_systems/download/)
* Document that a recent CMake needs to be installed on Ubuntu 20.04, using Kitware Apt repo.
* On Windows it is now necessary to source the `setup.bat` after the cmake invocation and before the build.

This PR also adds a `ROBOTOLOGY_SUPERBUILD_BUILD_SEPARATE_YARP_ROS` CMake option, that is `ON` by default when used with `ROBOTOLOGY_PROJECT_TAGS` set to `Unstable`, but that can be manually enable if users want to set `YCM_EP_DEVEL_MODE_YARP` to `ON` and manually switch to use `YARP` master and the future YARP 3.10. 

fyi @SimoneMic @GiulioRomualdi @S-Dafarra

PR #1630, opened 2024-04-02
Drop Debian 10 Buster support

traversaro:
This will be required to support YARP 3.10, as otherwise we do not have easy and maintainable way of installing CMake in the version that will be required by YARP 3.10.

This was added in https://github.com/robotology/robotology-superbuild/pull/1547, but I separate it in a separate PR to better visibility. 

**IMPORTANT: This change will make future robotology-superbuild releases not compatible with pc104-based iCub robots, as the last available image for pc104 is based on Debian 10 Buster, see https://icub-tech-iit.github.io/documentation/icub_operating_systems/pc104/icub-live/#versioning .**

As Debian 10 uses glibc 2.28 (see https://repology.org/project/glibc/versions), it should be possible to install software with conda-forge dependencies on Debian 10 for some time in the future.

@Nicogene @valegagge @pattacini @maggia80

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,6 @@ jobs:
         docker_image:
           - "ubuntu:focal"
           - "ubuntu:jammy"
-          - "debian:buster-backports"
           - "debian:bullseye"
 
         project_tags:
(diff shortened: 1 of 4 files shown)


pattacini (approved):

traversaro:
I wanted to cleanup the part related to `libdc1394-dev`, but apparently on Ubuntu 20.04 this is not possible, we need to restore the old script.

Nicogene (approved):

PR #1653, opened 2024-05-16
Suggest to use conda and miniforge instead of mamba and mambaforge

traversaro:
Back in time when we started using conda (~2020/2021?), the conda tool was extremely slow, and that is the reason why mamba was developed, and why we suggested everyone to use mamba. Similarly, we suggested to install mambaforge as a distribution instead of miniforge3, as mambaforge contained by default. See https://github.com/robotology/robotology-superbuild/pull/847 for the related PR.

Since those days, a lot of changed. conda has  been updated to use the mamba solver, and so it has become much faster and more and less as fast as mamba . This is true that as long as you have a recent enough `conda`, to verify that check that the commands conda config --show solver returns solver: libmamba, see https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community .

For this reason, several things has started to change:

* miniforge3 started installing mamba as well, so miniforge3 is the installer suggested by the conda-forge community, while mambaforge use is discouraged (see the README of https://github.com/conda-forge/miniforge for more info)
* mamba starts to have subtle bugs (such as https://github.com/mamba-org/mamba/issues/3222) with respect to  fully supported tools like pixi and conda
* users have been historically confused by the dualism between conda and mamba (and even worse micromamba, but fortunatly we never documented its use too much).

So, also in view of the conda/pixi training that I eventually need to do and to reduce the confusion of users, I have the intention to start recommending the use of conda (in place of mamba) and miniforge3 (in place of mambaforge) everywhere in our docs, and this PR is the first step in that direction.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,12 +13,12 @@ on:
     types: [published]
   schedule:
   # * is a special character in YAML so you have to quote this string
-  # Execute a "nightly" build at 2 AM UTC 
+  # Execute a "nightly" build at 2 AM UTC
   - cron:  '0 2 * * *'
 
 
 jobs:
-  conda-build:    
+  conda-build:
     name: '[conda:Tags:${{ matrix.project_tags }}@${{ matrix.os }}@${{ matrix.build_type }}]'
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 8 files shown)


Nicogene (approved):

traversaro:
generate-conda-packages CI job: https://github.com/robotology/robotology-superbuild/actions/runs/9109800751 .

xela-95 (approved):

traversaro:
> matlab-one-line-install CI job: https://github.com/robotology/robotology-superbuild/actions/runs/9109828346 (as this PR modified the one line matlab install script)

new run: https://github.com/robotology/robotology-superbuild/actions/runs/9109938233 .

martinaxgloria (approved):

Issue #1563, opened 2024-01-05
Do not add anymore <install_prefix>\share\gazebo\models to GAZEBO_MODEL_PATH ?


traversaro:
See https://github.com/robotology/gz-sim-yarp-plugins/issues/45#issuecomment-1879012926 .

PR #1655, opened 2024-05-28
Add generation of non-periodic conda recipes

traversaro:
This PR adds an action to generate some non-periodic conda recipes that are independent from the actual packages of the robotology superbuild. The main goal of such packages is to create metapackages that are useful for configuration.

The first example of such meta packages is `vvv-school-activation`, a package that is just empty and is only useful to set some environment variables used by the [`vvv-school`](https://github.com/vvv-school) and [`easy-peasy-robotics`](https://github.com/easy-peasy-robotics) courses.

This will fix the following issues:
* Fix https://github.com/robotology/robotology-superbuild/issues/1563
* Fix https://github.com/vvv-school/tutorial_gazebo-simple-model/issues/1


traversaro:
fyi @xela-95 , merging as it is required by some internal training.

Issue #1083, opened 2022-04-05
Single-Config Ninja build fails on Windows April 2022


traversaro:
I tried to build locally with ninja, and it is failing with error:
~~~
-- Long integers (64bit) are ON
-- Configuring done
-- Generating done
CMake Error:
  Running

   'C:/Users/STraversaro/AppData/Local/mambaforge/envs/robsub/Library/bin/ninja.exe' '-C' 'C:/src/robotology-superbuild/build/src/osqp' '-t' 'recompact'

  failed with:

   ninja: error: build.ninja:320: multiple rules generate out/osqp.lib [-w dupbuild=err]





CMake Generate step failed.  Build files cannot be regenerated correctly.
ninja: build stopped: subcommand failed.
~~~

traversaro:
Probably related to https://github.com/tesseract-robotics/trajopt/pull/287 .

traversaro:
Apparently Windows-rebuild of the superbuild takes a lot of time. Ideally solving this and https://github.com/conda-forge/vc-feedstock/issues/39 and documenting the ninja based workflow could drastically improve this.

traversaro:
> Apparently Windows-rebuild of the superbuild takes a lot of time. Ideally solving this and [conda-forge/vc-feedstock#39](https://github.com/conda-forge/vc-feedstock/issues/39) and documenting the ninja based workflow could drastically improve this.

Today I did for debug a build with `ROBOTOLOGY_ENABLE_DYNAMICS_FULL_DEPS` and I took literally forever, I think it is quite important to fix this.

PR #1682, opened 2024-08-10
Fix support for building on Windows with Ninja

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1083 .

To do so, it was necessary to fix support for building shared libraries and disabling static libraries in osqp 0.6.3, this was already done since a lot of time in osqp master, but as long as we do not get a release of osqp, we continue to use osqp 0.6.3 . The related commits are:
* https://github.com/robotology-dependencies/osqp/commit/060751ffcca0d0f2b800baf09b2e067310d9ddb2
* https://github.com/robotology-dependencies/qdldl/commit/3770026c0546216df29c0ea37102df20d33f2ba5

### DIFF
--- a/cmake/Buildosqp.cmake
+++ b/cmake/Buildosqp.cmake
@@ -12,7 +12,7 @@ ycm_ep_helper(osqp TYPE GIT
               FOLDER src
               CMAKE_ARGS -DUNITTESTS:BOOL=OFF
                          -DOSQP_BUILD_STATIC_LIB:BOOL=OFF
-                         -DQDLDL_BUILD_STATIC_LIB:BOOL=OFF))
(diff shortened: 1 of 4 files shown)


traversaro:
For now I will not update the docs, as anyhow ninja + superbuild has its own problems (see https://discourse.cmake.org/t/efficiency-issues-with-externalproject-sub-builds-under-ninja and https://gitlab.kitware.com/cmake/cmake/-/issues/22145), but it may be useful to know this for faster builds on Windows, fyi @S-Dafarra @mebbaid or anyone else that actually uses the superbuild on Windows (not sure who).

traversaro:
At least in CI, Ninja builds take ~half of the time of Visual Studio builds.

Issue #1439, opened 2023-07-06
Drop Ubuntu 20.04 support for apt dependencies


traversaro:
I know that we did not fully even solved https://github.com/robotology/robotology-superbuild/issues/481. Anyhow the problems related to Ubuntu 20.04  apt dependencies are increasting, so I wanted to start opening this issue to start collecting things that block us from updating to 20.04:

* [x] JetPack SDK stuck to Ubuntu 20.04 
* [ ] iCubOS not running on Ubuntu 20.04 
* [x] AMI machine(s) not updated to Ubuntu 22.04

@Nicogene do you know the situation of icub-tech's machines and Docker images?
@xEnVrE do you know about the situation of HSP's machine or other constraints (like in https://github.com/robotology/robotology-superbuild/issues/481 we had CUDA-related problems).

xEnVrE:
Hi @traversaro,

at the moment all the machines (personal and those of iCubGenova11) are mostly using Ubuntu 20.04 - but I counted at least 3 machines already using Ubuntu 22.04. 

I would say we would need to update at least those of iCubGenova11 when the support is removed. I cannot comment on the machines of the R1 setup - not sure they are using the superbuild.

As regards applications w/ CUDA they mostly run on `docker` where we have, I think, a mix of situations (ranging from Ubuntu 18.04 to Ubuntu 22.04) and not all of them require the software provided by the superbuild.

Nicogene:
> @Nicogene do you know the situation of icub-tech's machines and Docker images?

@valegagge can confirm it but I think that the docker images are focal-based:

https://github.com/icub-tech-iit/docker-deployment-images/blob/2961c08bc4fc39c75e03adbf2d5371fddcb2723b/dockerfile_images/basic/superbuild/conf_build.ini#L2

valegagge:
Hi @Nicogene ,

> @valegagge can confirm it but I think that the docker images are focal-based:

yes!I confirm. 
When we are ready to move to the 22 we can start to update the docker images and our laptops.

traversaro:
Possibly related PR: https://github.com/robotology/yarp/pull/2880 .

traversaro:
[...]

PR #1684, opened 2024-08-11
Add cli11 dependency

traversaro:
This dependency is quite small, and will be required by https://github.com/robotology/gz-sim-yarp-plugins/pull/179 . For conda/pixi I just add it as any required dependency (and anyhow it is already installed as gz-sim transitive dependency) while for apt I documented its installation where we explain how to install gz-sim as `libcli-dev` is not available in Ubuntu 20.04 (https://github.com/robotology/robotology-superbuild/issues/1439). Once we drop the Ubuntu 20.04 installation, we can just add it to the `apt.txt` file.


pixi.toml
```diff
@@ -79,6 +79,7 @@ ace = "*"
 asio = "*"
 assimp = "*"
 boost = "*"
+cli11 = "*"
```
xela-95, line 82: Shouldn't this be placed under the "ROBOTOLOGY_USES_GZ" dependencies for clarity?

xela-95 (approved):
Thank @traversaro! Just a minor non blocking-comment.

LGTM!

pixi.toml
traversaro: Good point. To be honest, as it is quite a small library, I prefer to keep it in the group of generic dependencies, so that if other libraries start depend on it I do not need to remember to move it.

xela-95: Ok thanks!

PR #1693, opened 2024-08-14
Remove pin of YCM

traversaro:
It should not be necessary after https://github.com/robotology/ycm-cmake-modules/pull/458 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,10 +269,10 @@ jobs:
       fail-fast: false
       matrix:
         build_type: [Release]
-        os: [ubuntu-20.04, windows-2019]
+        os: [ubuntu-22.04, windows-2019]
         project_tags: [Default, Unstable, LatestReleases]
         include:
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


traversaro:
Argh, we now have again:

~~~
CMake Error at build/_deps/ycm-src/cmake-next/proposed/ExternalProject.cmake:3223 (_ep_parse_arguments):
  _ep_parse_arguments Function invoked with incorrect arguments for function
  named: _ep_parse_arguments
  build/_deps/ycm-src/modules/YCMEPHelper.cmake:1113 (ExternalProject_Add)
  cmake/Buildhuman-gazebo.cmake:7 (ycm_ep_helper)
  build/_deps/ycm-src/modules/FindOrBuildPackage.cmake:205 (include)
  cmake/RobotologySuperbuildLogic.cmake:151 (find_or_build_package)
  CMakeLists.txt:67 (include)
~~~

In Ubuntu 20.04 GitHub Actions images. However, I think that is a particularly strange case as in that image there are multiple CMake installed with different versions. I think we can drop those CI tests, and just keep the one done in 20.04 with Docker.

PR #1697, opened 2024-08-22
pixi: Lower minimum linux kernel version required to 4.18

traversaro:
The default version of pixi 0.27.1 is `5.10`, and so running anything on a HPC system running Rocky Linux 8 results in errors like:

~~~
  × The current system has a mismatching virtual package. The project requires '__linux' to be at least version '5.10' but the system has version '4.18.0'
~~~

xref: https://github.com/prefix-dev/pixi/issues/1884


traversaro:
This is not necessary anymore if you use pixi 0.28, see https://github.com/prefix-dev/pixi/releases/tag/v0.28.0 .

PR #1723, opened 2024-09-25
Cleanup Python snippet required to run Python bindings on Windows

traversaro:
I modified the code to fail with no effect on non-Windows and non-superbuild installation, so that the snippet can be added to any script without preventing to run the script in context different from a Windows superbuild install.

### DIFF
--- a/doc/cmake-options.md
+++ b/doc/cmake-options.md
@@ -372,9 +372,8 @@ Furthermore, due to Python ignoring the directories in `PATH`, before running py
 
 ~~~python
 import os
-import platform
 
-if platform.system() == "Windows":
+if os.name == "nt":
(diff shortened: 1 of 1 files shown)


doc/cmake-options.md
```diff
@@ -372,7 +372,12 @@ Furthermore, due to Python ignoring the directories in `PATH`, before running py
 
 ~~~python
 import os
-os.add_dll_directory(os.path.join(os.environ['ROBOTOLOGY_SUPERBUILD_INSTALL_PREFIX'], 'bin'))
+import platform
+
+if platform.system() == "Windows":
```
giotherobot, line 377: You can do this without additional import with `if os.name == "nt":`

traversaro: Thanks, applied in https://github.com/robotology/robotology-superbuild/pull/1723/commits/a0ab067602aaf7761e73a68b7f417aed4c7e6925 .

giotherobot (approved):

PR #1727, opened 2024-10-01
Migrate icub-firmware-shared, icub-main and human-dynamics-estimation conda packages to be installed via conda-forge

traversaro:
To mark the change, bump `CONDA_BUILD_NUMBER` to 140.


traversaro:
The only faiure is in the `robotology-distro-all` metapackage: https://github.com/robotology/robotology-superbuild/actions/runs/11126270082 , I think we can merge.

PR #1728, opened 2024-10-03
Enable ROBOT_TESTING under Apple Silicon macOS

traversaro:
See if https://github.com/robotology/blocktest/issues/66 is still present.


traversaro:
> See if https://github.com/robotology/blocktest/issues/66 is still present.

Indeed the problem was solved! Let's merge.

PR #1734, opened 2024-11-09
Update YCMBootstrapFetch.cmake to avoid CMP0169 related warnings and update YCM to 0.17.1

traversaro:
This will remove all the noisy warnings:

~~~
CMake Warning (dev) at D:/miniforge3/envs/robsub/Library/share/[cmake-3](https://gitlab.kitware.com/cmake/cmake/-/issues/3).30/Modules/FetchContent.cmake:1953 (message):
Calling FetchContent_Populate(YCM) is deprecated, call
FetchContent_MakeAvailable(YCM) instead. Policy CMP0169 can be set to OLD
to allow FetchContent_Populate(YCM) to be called directly for now, but the
ability to call it with declared details will be removed completely in a
future version.
cmake/YCMBootstrapFetch.cmake:103 (FetchContent_Populate)
CMakeLists.txt:60 (include)
~~~

in recent CMake releases.

### DIFF
--- a/cmake/YCMBootstrapFetch.cmake
+++ b/cmake/YCMBootstrapFetch.cmake
@@ -97,10 +97,4 @@ FetchContent_Declare(YCM
                      GIT_REPOSITORY https://github.com/${YCM_FETCHCONTENT_REPOSITORY}
                      GIT_TAG ${YCM_FETCHCONTENT_TAG})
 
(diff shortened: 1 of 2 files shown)


traversaro:
This highlighted a bug that was fixed in https://github.com/robotology/ycm-cmake-modules/pull/464 .

PR #1738, opened 2024-11-11
Update Buildrobometry.cmake after the addition of IRawValuesPublisher

Nicogene:
In `telemetryDeviceDumper` it has been added the usage of `IRawValuesPublisher` by @MSECode, this made the device dependent also by ICUB 2.7.0 or higher.

I am unsure if the superbuild compiles the device right now, because if the dependencies are not found the compilation of the device is skipped.

See:
- https://github.com/robotology/robometry/pull/190

cc @traversaro @S-Dafarra @valegagge

### DIFF
--- a/cmake/Buildrobometry.cmake
+++ b/cmake/Buildrobometry.cmake
@@ -5,6 +5,7 @@
 include(YCMEPHelper)
 
 find_or_build_package(YARP QUIET)
+find_or_build_package(ICUB QUIET)
 find_or_build_package(matioCpp QUIET)
 
 ycm_ep_helper(robometry TYPE GIT
@@ -15,6 +16,7 @@ ycm_ep_helper(robometry TYPE GIT
               FOLDER src
               DEPENDS YCM
                       YARP
+                      ICUB
                       matioCpp)
 
 set(robometry_CONDA_PKG_NAME librobometry)


traversaro:
> I am unsure if the superbuild compiles the device right now, because if the dependencies are not found the compilation of the device is skipped.

Not a big problem, but is not ideal, as if for any reason the dependencies are not found the requested device is silently not compiled instead of raising an error. 

What I typically do is to have an option that users can be enabled or disabled, and set its default value based on the output of the optional `find_package` calls, something like:

~~~cmake
set(ROBOMETRY_COMPILE_TELEMETRYDEVICEDUMPER_DEFAULT_VALUE OFF)
if(NOT DEFINED ROBOMETRY_COMPILE_TELEMETRYDEVICEDUMPER)
    find_package(iCubDev 2.7.0 QUIET)
    set(ROBOMETRY_COMPILE_TELEMETRYDEVICEDUMPER_DEFAULT_VALUE ${iCubDev_FOUND})
endif()
option(ROBOMETRY_COMPILE_TELEMETRYDEVICEDUMPER "If ON, compile telemetryDeviceDumper YARP device (requires YARP and ICUB)" ${ROBOMETRY_COMPILE_TELEMETRYDEVICEDUMPER_DEFAULT_VALUE })
if(ROBOMETRY_COMPILE_TELEMETRYDEVICEDUMPER)
    find_package(iCubDev 2.7.0 REQUIRED)
    add_subdirectory(telemetryDeviceDumper)
else()
~~~

traversaro (approved):

traversaro:
Having said that, the change seems to work, let's merge.

Issue #1740, opened 2024-11-13
Update YARP in ROBOTOLOGY_PROJECT_TAGS=Stable to yarp-3.10


traversaro:
Ideally we should try to do this before the 2024.11.0 distro. However, we may want to run some tests on the real robots before doing that. I will coordinate the tests on the real robots on the AMI side, @Nicogene can you handle this on the iCubTech side if you think you could do some tests on the real robot?

Issue #1348, opened 2023-02-16
Add ROBOTOLOGY_USES_ROS2 option


traversaro:
With this option enable in CI, we would have catched https://github.com/ami-iit/bipedal-locomotion-framework/issues/606 before it affected some users.

traversaro:
Related to our discussion of today. @Nicogene @randaz81 @pattacini @martinaxgloria @elandini84

traversaro:
Related to this, as we need a version of yarp-devices-ros2 compatible with the default configuration of the superbuild, I created one in  https://github.com/ami-iit/yarp-devices-ros2/tree/yarp-3.9 .

PR #1746, opened 2024-11-26
Add `ROBOTOLOGY_USES_ROS2` and `ROBOTOLOGY_USES_MOVEIT` options

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1348 .

This is still preliminary work, and it only works if `-DROBOTOLOGY_PROJECT_TAGS` is set to `Unstable` until we fix https://github.com/robotology/robotology-superbuild/issues/1740 . However as it is working fine locally, I thought it was worth opening the PR to collect preliminary feedback. 

For now I still need to add CI for these two options, but to test that the options are working I run:

~~~
pixi run build-ros2
~~~

I did not check if anything is working, but at least I could check that the ROS packages are correctly installed via `pixi run build-ros2 && pixi run -e ros2 ros2 pkg list`.

Note that the documentation is provided to enable the options with all supported way of installing dependencies (`apt`, `conda`, `pixi`), I just tested the option with `pixi` as it is faster.

This PR is also requireds https://github.com/icub-tech-iit/xcub-moveit2/pull/27 to be merged and released.


martinaxgloria:
> @martinaxgloria @Nicogene how would you like the (future, it will not be fast) conda package to be named? `xcub-moveit` or `xcub-moveit2`?

I'd like `xcub_moveit2` as the name of the repository. I think that xcub_moveit could be misleading since someone could think that it is a MoveIt 1 project. But I'm also open to name the conda package without the moveit version specified

traversaro:
This PR is missing just a tag for xcub-moveit2 and then we are ready to go. @martinaxgloria can you review the session on the required dependencies for apt in https://github.com/robotology/robotology-superbuild/pull/1746/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R151 . For the conda-forge ones I tested it in CI, but for apt I do not plan to add CI for now (too much effort in mantainance).

martinaxgloria:
Hi @traversaro, everything seems ok to me, thanks

traversaro:
The Windows ROS2 CI fails due to the length of the build path. There are several possible workarounds, but to unblock the build let's skip the CI for now.

PR #1770, opened 2024-12-16
Fix ycm broken links in README.md

Nicogene:
As per title


traversaro:
Thanks! The CI is a christmas tree but it is probably unrelated.

Issue #1773, opened 2024-12-17
conda CI great 2024 christmas failure


traversaro:
We had a bit of perfect storm of CI failures for conda, probably related to the fact that all the dependencies are now available with Python 3.13 . 

The issues are:
* `pip check` now fails with `sip 6.7.12 requires setuptools, which is not installed.`
* On Windows, installing all dependencies takes ~1 hour and then fails.

traversaro:
>  are now available with Python 3.13 .

Actually Python 3.10 is still installed.

traversaro:
Actually the before/after difference is quite minimal:

~~~diff
--- <unnamed>
+++ <unnamed>
@@ -49,7 +49,6 @@
 cached-property           1.5.2                hd8ed1ab_1    conda-forge
 cached_property           1.5.2              pyha770c72_1    conda-forge
 cairo                     1.18.0               h3faef2a_0    conda-forge
-certifi                   2024.8.30          pyhd8ed1ab_0    conda-forge
 cfitsio                   4.4.0                hbdc6101_1    conda-forge
 cli11                     2.4.2                h5888daf_0    conda-forge
 cmake                     3.31.2               h74e3db0_1    conda-forge
@@ -321,8 +320,8 @@
 lz4-c                     1.9.4                hcb278e6_0    conda-forge
 lzo                       2.10              hd590300_1001    conda-forge
 make                      4.4.1                hb9d3cd8_2    conda-forge
-matplotlib                3.9.4           py310hff52083_0    conda-forge
-matplotlib-base           3.9.4           py310h68603db_0    conda-forge
+matplotlib                3.10.0          py310hff52083_0    conda-forge
+matplotlib-base           3.10.0          py310h68603db_0    conda-forge
 matplotlib-inline         0.1.7              pyhd8ed1ab_1    conda-forge
 mesa-khr-devel-cos7-x86_64 18.3.4            ha675448_1106    conda-forge
 mesa-libgl-cos7-x86_64    18.3.4            ha675448_1106    conda-forge
@@ -366,7 +365,7 @@
 pexpect                   4.9.0              pyhd8ed1ab_1    conda-forge
 pickleshare               0.7.5           pyhd8ed1ab_1004    conda-forge
 pillow                    10.3.0          py310hf73ecf8_0    conda-forge
-pip                       24.3.1             pyh8b19718_0    conda-forge
+pip                       24.3.1             pyh145f28c_1    conda-forge
[...]

PR #1774, opened 2024-12-17
Refresh conda dependencies December 2024

traversaro:
Hopefully this will fix https://github.com/robotology/robotology-superbuild/issues/1773, a few different fixes ended up here:
* Remove the ffmpeg pin to 6 added in https://github.com/robotology/robotology-superbuild/pull/1652 as we now switched to YARP 3.10 supports ffmpeg 7
* Remove the use of CDTs to depend on opengl, and instead use `libgl-devel` as now it is documented in https://conda-forge.org/docs/maintainer/maintainer_faq/#how-do-i-fix-the-libglso1-import-error (see https://github.com/ami-iit/yarp-device-openxrheadset/pull/57) for more explanations
* Replace deprecated `boost` package with `libboost-devel`
* As the pixi dependencies were changed, we regenerated the pixi lock file


traversaro:
The failures are all related to https://github.com/conda-forge/pip-feedstock/issues/132, so we can merge as anyhow this is an improvement.

PR #1775, opened 2024-12-17
Add advanced ROBOTOLOGY_SKIP_ROBOTS_CONFIGURATION CMake option

traversaro:
This is option is not meant to be used in general, but it can be useful in setups where there is a single  robots-configuration installed outside of the robotology-superbuild that is shared by different robotology-superbuild, that are build in different environments or container images.

This option is meant to be used in specific contexts, so it will be documented there and is not documented (and it is marked as advanced, to be hidden by default) in the `robotology-superbuild`.

This option is meant to be used with the functionality added to robots-configuration in https://github.com/robotology/robots-configuration/pull/697 .


S-Dafarra:
I wonder if it makes sense to have a dummy ``Config.cmake`` for robots-configuration that simply sets ``FOUND`` to true

S-Dafarra (approved):

traversaro:
> I wonder if it makes sense to have a dummy `Config.cmake` for robots-configuration that simply sets `FOUND` to true

Even if that was there, I am not sure if it useful, for the following reasons:
* We would need to add the install folder of robots-configuration to `CMAKE_PREFIX_PATH` (but this is doable)
* More importantly, despite the name `find_or_build_package` does not try to call `find_package` since a long time ago (I can't even find the PR), as that behavior was only an endless source of problems of users for which the robotology-superbuild was breaking as they were finding an old or wrongly configured version of YARP/iDynTree/etc/etc, while since we started ignoring system installed packages, I can't remember a single time in which this behaviour was a problem or a limitation (until this comment, at least : ) ).

traversaro:
> * since a long time ago (I can't even find the PR)

PR found: https://github.com/robotology/robotology-superbuild/pull/389 .

S-Dafarra:
Clear! At this point, maybe it makes sense to add the definition of an env variable indicating where are located the configurations files. This would simplify the definition of the additional commands we currently use on the robot (see for example https://github.com/robotology/robots-configuration/blob/6b22848d0654505ea4277e0c6d109dea0c58509b/ergoCubSN000/extra/scripts/installRobotsConfiguration.sh#L11). Related comment: https://github.com/robotology/robots-configuration/pull/698#discussion_r1888745941

If you agree, I can also provide a PR for the ``setup.sh``/``.bat`` files

traversaro:
> Clear! At this point, maybe it makes sense to add the definition of an env variable indicating where are located the configurations files. This would simplify the definition of the additional commands we currently use on the robot (see for example https://github.com/robotology/robots-configuration/blob/6b22848d0654505ea4277e0c6d109dea0c58509b/ergoCubSN000/extra/scripts/installRobotsConfiguration.sh#L11). Related comment: [robotology/robots-configuration#698 (comment)](https://github.com/robotology/robots-configuration/pull/698#discussion_r1888745941)

Sure, that probably should be placed in `.bashrc_ergoCub`/ `.bashrc_iCub` ?

traversaro:
I also cherry-picked the change in the `releases/2024.11`, so it will end up in a possible 2024.11.2 release, see https://github.com/robotology/robotology-superbuild/commit/568b758ea8b57b4c8e6f9ae1c5233a5a2549a7bf .

PR #1792, opened 2025-02-05
Change depends_on with depends-on in pixi.toml

traversaro:
The `depends-on` syntax was supported since pixi 0.21 (see https://github.com/prefix-dev/pixi/pull/1310), so there risk of creating troubles to any user with an old pixi is non existent, as anyhow the lock file version that we used is newer if I am not wrong.

This should reduce the amount of warnings printed on the terminal of @AntonioViscomi .


traversaro:
@AntonioViscomi if you like the change feel free to review the PR (see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews), thanks!

AntonioViscomi (approved):
Thank you, warnings are no longer displayed!

PR #1793, opened 2025-02-05
pixi: Use a single CMAKE_BUILD_TYPE env variable to set the build type

traversaro:
@AntonioViscomi was reporting some problem when trying to switch some subprojects of the superbuild to `RelWithDebInfo`. By looking at the pixi.toml, I noticed that the information on the CMake build type was duplicated everywhere, while probably it is simpler to just set once the [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) env variable.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -12,6 +12,11 @@ platforms = ["linux-64", "linux-aarch64", "win-64", "osx-64", "osx-arm64"]
 [system-requirements]
 linux = "4.18"
 
+[activation.env]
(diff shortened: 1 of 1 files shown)


AntonioViscomi (approved):
The build is not overwriting the cmake options anymore. Thanks!
```bash
C:\Users\aviscomi\robotology-superbuild>pixi run build -j16
✨ Pixi task (configure in default): cmake -G Ninja -S . -B .build
```
cc @traversaro

PR #1818, opened 2025-03-03
Use latest conda and conda-libmamba-solver to avoid segfaults on Windows

traversaro:
In the last months we had frequengly segfaults when using conda install on Windows. This CI upgrades the used version of mamba to 2.*, to check if that fixes the segfaults.


traversaro:
I am not sure if it was due to the change or we were just lucky, but there was not segfault in this speicfic CI run. To check if this actually works in all cases, let's merge.

Issue #1854, opened 2025-05-23
Upgrade to YARP 3.12.0


traversaro:
YARP 3.12.0 will be released in a few days by branching out of the YARP's `master` branch. 

Unfortunately, I forgot that in https://github.com/robotology/robotology-superbuild/pull/1807 I had pinned the YARP version in `Unstable` branches to 3.11.1, so a whole ~3 months went on without the CI checking for CMake and/or C++ breaking changes in YARP, so we need do to that as soon as possible.

Beside C++ or CMake breaking changes, that we can easily handle with conditional on YARP version, there was also a rename of the (widely used) device `opencv_grabber` to `openCVgrabber` in https://github.com/robotology/yarp/pull/3225 . 
The `opencv_grabber` is widely used, at least in the AMI team, whenever a webcam is used to save data (typically as part of a setup used to save the video of an experiment as an external view, see https://github.com/search?q=org%3Aami-iit+opencv_grabber&type=code&p=1).

The rename of a YARP device is tricky as:
* It typically only catched at run time, so we don't catch regressions in CI
* The device name is typically embedded in a .xml yarprobotinterface file, so it is not easy to change the name of the device name to use based on the YARP version used

Possible way of dealing with that include:
* **SOL1**: We wait for the release for YARP 3.12.0, we update all the projects (public and private) that use `opencv_grabber` to `openCVGrabber`, and then we tag all of them, so we bump their version in LatestReleases all together. I am not a big fan of this, as them many user may want to use new projects with older YARP version (for example, when dealing with a robot setup that is shared and they do not control), so I typically prefer for project to support multiple YARP versions at the same time
* **SOL2**: We provide somehow in the superbuild a `opencv_grabber` device that just redirects to `openCVGrabber` printing a warning, and then we migrate in some time to `openCVGrabber` when we are sure that all ergoCub robots have been migrated to `openCVGrabber`

I need to investigate the feasibility of **SOL2**, but if that is technically possible I would probably go for that solution.

fyi @S-Dafarra @giotherobot @isorrentino @carloscp3009 @GiulioRomualdi @Giulero @HosameldinMohamed @xela-95 @LoreMoretti (tagging a few person across AMI team that I know that they use logger setups that use `opencv_grabber`).

S-Dafarra:
I think that mostly this affects the BLF logger files. I believe this should not be too hard to handle since the logger files are robot dependent. So I guess that if we keep the robot and the logger files aligned we should be fine

[...]

PR #1855, opened 2025-05-23
Switch YARP in ProjectsTagsUnstable back to master

traversaro:
First step of https://github.com/robotology/robotology-superbuild/issues/1854 . YARP was pinned to 3.11.2 in https://github.com/robotology/robotology-superbuild/pull/1807, so we need to switch back YARP in unstable branches to YARP master branch, and solve any possible incompatibility.


traversaro:
Ok, CI is happy, let's merge to quickly see when `master` of YARP breaks again.

Issue #1503, opened 2023-10-17
Use proxsuite when using the superbuild


xEnVrE:
We need to use `proxsuite` for a project involving robots for which we also use the `robotology-superbuild`. Given that we were not aware of the fact that the `superbuild` fetches and installs that library, we have cloned `proxsuite` and installed on the system..

We then realized that something was wrong as our `CMake` project could not find the `proxsuite::proxsuite-vectorized` target. After some investigation we discovered that our project was indeed try to link against the `proxsuite` library installed by the `superbuild` which however is compiled with `-DBUILD_WITH_VECTORIZATION_SUPPORT:BOOL=OFF` to support `Ubuntu 20.04`.

Could you give us a suggestion for this scenario? At the moment we are forcing `CMake` to use our installation, built with vectorization, by exporting `proxsuite_DIR` differently.

Thanks

cc @traversaro @PasMarra

traversaro:
Can we have a quick chat today to understand the context on this? Thanks! @xEnVrE @PasMarra

xEnVrE:
For sure!

traversaro:
We dropped Ubuntu 20.04 support with apt dependencies in https://github.com/robotology/robotology-superbuild/pull/1802, I guess now we can just add back simde dependency and `-DBUILD_WITH_VECTORIZATION_SUPPORT:BOOL=ON` .

PR #1861, opened 2025-05-31
Enable BUILD_WITH_VECTORIZATION_SUPPORT in proxsuite build

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1503 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,7 @@ jobs:
       shell: bash -l {0}
       run: |
         # Dependencies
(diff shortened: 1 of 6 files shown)


Nicogene:
> We can also drop this. The request fom this came from Nicola Piga, I am not sure if there is any interest on this.

The main issue is to update the .zip of vcpkg dependencies, maybe @PasMarra is interested in this feature?

Otherwise yes we can just drop it

PasMarra:
Hi @Nicogene, no problem at all. You can just drop it.

Issue #1497, opened 2023-10-11
conda packages generated by the robotology-superbuild should prepend values to existing environment variables, not append


traversaro:
For a long time, in all activation scripts I always "appended" values to existing environment variables. My reasoning was to be "kind" with the environment. However, in the case of conda packages, this does not make sense: if there is something defined at the system level, the latest activated environment should have the precedence, not the first.

cc @Nicogene @xEnVrE

Nicogene:
cc @martinaxgloria

PR #1892, opened 2025-08-20
Cleanup unused multisheller script and prepend variables instead of appending

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1497 .


martinaxgloria (approved):
Thanks!

PR #1893, opened 2025-08-20
Fix conda package generation

traversaro:


traversaro:
We also need https://github.com/robotology/speech/pull/46 .

Issue #1908, opened 2025-10-20
CI Failure October 2025


traversaro:
The failure is:

~~~

2s
Run source ./b/install/share/robotology-superbuild/setup.sh
robot-log-visualizer 0.0.0 requires pyside2, which is not installed.
Error: Process completed with exit code 1.
Run source ./b/install/share/robotology-superbuild/setup.sh
robot-log-visualizer 0.0.0 requires pyside2, which is not installed.
Error: Process completed with exit code 1.
~~~

this is a regression of https://github.com/ami-iit/robot-log-visualizer/pull/104 . @xela-95 experienced problems with `pyside2` so we want to consider reverting that (at least let `qtpy` use `pyqt5` instead of `pyside2`) but for the time being the easist option is just to install also `pyside2` and `qtpy` in the superbuild deps.

PR #1909, opened 2025-10-20
Add qtpy and pyside2 dependency to robot-log-visualizer

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1908 .

fyi @xela-95 @GiulioRomualdi @Giulero


traversaro:
I think the CI is now failing as conda is down due to the global AWS outage.

Issue #1915, opened 2025-10-31
Failed to open GUIs in clean Windows installation


S-Dafarra:
After a clean installation on Windows, if I try to launch ``yarpview`` I get the following error

<img width="617" height="238" alt="Image" src="https://github.com/user-attachments/assets/4099ddc4-6e41-45e6-b6d8-04aabd1cec2f" />

```
The application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem
```

<details>
<summary>Here the <code>conda list</code></summary>

```
IITBMP014LW025 D:\
robsub ❯ conda list
# packages in environment at D:\miniforge3\envs\robsub:
#
# Name                    Version                   Build  Channel
_libavif_api              1.3.0                h57928b3_2    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
ace                       8.0.5                h7034029_0    conda-forge
aiohappyeyeballs          2.6.1              pyhd8ed1ab_0    conda-forge
aiohttp                   3.13.1          py311he1a1d09_0    conda-forge
aiosignal                 1.4.0              pyhd8ed1ab_0    conda-forge
ampl-asl                  1.0.0                he0c23c2_2    conda-forge
aom                       3.9.1                he0c23c2_0    conda-forge
asio                      1.29.0               h63175ca_0    conda-forge
assimp                    5.4.3                hfc39600_1    conda-forge
asttokens                 3.0.0              pyhd8ed1ab_1    conda-forge
attrs                     25.4.0             pyh71513ae_0    conda-forge
blosc                     1.21.6               hfd34d9b_1    conda-forge
brotli                    1.1.0                hfd05255_4    conda-forge
brotli-bin                1.1.0                hfd05255_4    conda-forge
bzip2                     1.0.8                h0ad9c76_8    conda-forge
ca-certificates           2025.10.5            h4c7d964_0    conda-forge
cached-property           1.5.2                hd8ed1ab_1    conda-forge
cached_property           1.5.2              pyha770c72_1    conda-forge
cairo                     1.18.4               h5782bbf_0    conda-forge
catch2                    3.8.0                hc790b64_0    conda-forge
certifi                   2025.10.5          pyhd8ed1ab_0    conda-forge
cli11                     2.6.0                h5112557_0    conda-forge
cmake                     4.1.2                hdcbee5b_0    conda-forge
colorama                  0.4.6              pyhd8ed1ab_1    conda-forge
[...]

PR #1916, opened 2025-10-31
Fix name of qt.conf and qt6.conf files installed as a workaround

traversaro:
For some reason, in https://github.com/robotology/robotology-superbuild/pull/1853 we were installing the qt5 one (that is called `qt.conf`) as `qt6.conf` and the qt6 one (that is called `qt6.conf`) as `qt.conf`, so I suspect this may be the reason behind https://github.com/robotology/robotology-superbuild/issues/1915 .


traversaro:
The CI failure are unrelated, let's merge.

PR #1920, opened 2025-11-15
Add ccache support to pixi and pixi CI

traversaro:
This PR adds the support for using the `ccache` compilation cache tool in pixi, and ensure that the ccache cache is shared across CI builds, to reduce the time spent by CI in building everything.

The main advantage for this is that anyone that first build the superbuild locally, and then re-builds it from scratch, the build time is drastically reduced, as the build result are already contained in the cache. 

The project can be even downloaded in different folders, and the ccache can still be reused, thanks to the use of the `CCACHE_BASEDIR` env variable.

### DIFF
--- a/.github/workflows/pixi-ci.yml
+++ b/.github/workflows/pixi-ci.yml
@@ -47,8 +47,7 @@ jobs:
 
     - uses: prefix-dev/setup-pixi@v0.9.1
 
-    - name: ccache
-      uses: hendrikmuhs/ccache-action@v1.2
+    - uses: hendrikmuhs/ccache-action@v1.2
       with:
         key: ${{ matrix.os }}-${{ matrix.pixi_task }}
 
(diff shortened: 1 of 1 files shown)


pixi.toml
```diff
@@ -16,6 +16,10 @@ linux = "4.18"
 # This is a single place where CMAKE_BUILD_TYPE is defined, change its value here and
 # delete .build or .build-ros2 to change build type
 CMAKE_BUILD_TYPE = "Release"
+# ccache support
+CMAKE_C_COMPILER_LAUNCHER="ccache"
+CMAKE_CXX_COMPILER_LAUNCHER="ccache"
+CCACHE_BASEDIR="$PIXI_PROJECT_ROOT"
```
mjcarroll: FYI I think this ends up being a no-op on Windows, because you need to use the percent syntax:


```suggestion
CCACHE_BASEDIR="%PIXI_PROJECT_ROOT%"
```

PR #1942, opened 2026-01-12
Fix automatic update yarp-devices-ros2 and xcub-moveit2

### LINKED ISSUE #1941: Update latest releases does not update yarp-device-ros2
For some reason seems that the action that updates the latest release does not update `yarp-devices-ros2`

- https://github.com/robotology/robotology-superbuild/actions/workflows/update-latest-releases.yml

The latest is v2.0.0, latestlrelease.yml has v1.1.0

cc @martinaxgloria @MSECode

Nicogene:
It fixes #1941


traversaro:
I think this is a bug, can you remove the enable variables If they are not used?

PR #1946, opened 2026-01-14
Deprecate the support to gazebo classic, remove it from the CI

Nicogene:
This PR will depreacate the option `ROBOTOLOGY_USES_GAZEBO` in favour of `ROBOTOLOGY_USES_GZ`.

It will not be tested anymore in the CI and this option will be dropped in future releases.

cc @pattacini @randaz81 @PasMarra


Nicogene:
We are blocked by:
- #1947

PR #1981, opened 2026-05-12
Migrate the repositories to gbionics org

Nicogene:
cc @pattacini @valegagge @randaz81


Nicogene:
@traversaro I am not sure if we need to change also all the old distros  .yml, if someone checkout an old release, it will rely still on github redirect

traversaro:
> @traversaro I am not sure if we need to change also all the old distros .yml, if someone checkout an old release, it will rely still on github redirect

Yes, I think just changing the current one and the future ones is probably ok.

traversaro (approved):

Nicogene:
CI is green, merging

